PR #58, opened 2021-08-16
Revert "update package name to @hashicorp/ember-flight-icons"

Dhaulagiri:
Reverts hashicorp/flight#46

This change actually breaks the addon because references to the package need to also be updated. Let's revert to get things working and revisit this and ensure tests are passing before merging.


MelSumner (approved):
Thank you! @amyrlam and I found the same thing.

PR #63, opened 2021-08-16
Update in-repo app to include icon list

MelSumner:
If merged, this PR adds a link to the basic usage instructions and a rendered icon list to the in-repo app. 

image


Dhaulagiri:
> right now it feels really tedious

It _should_ be feasible for us to automatically generate each icon based on some array of inputs (either supplied as part of the icon export or we could reach into the sprite and pull them out I think)

amyrlam (approved):
pulled it locally, lgtm! 

it wasn't clear to me that we were doing a non-dynamic list as a first pass until the meeting, which is why i hadn't reviewed it before. 

it would be cool to have the instructions and https://github.com/hashicorp/flight/blob/main/ember-flight-icons/README.md be one source of truth in the future

amyrlam:
p.s. not sure why CI is stalled yet. i don't had admin privileges to override and merge. will investigate...

amyrlam:
my guess is this needs a rebase cause it doesn't show up in https://github.com/hashicorp/flight/actions if you can try that

MelSumner:
> It _should_ be feasible for us to automatically generate each icon based on some array of inputs (either supplied as part of the icon export or we could reach into the sprite and pull them out I think)

@Dhaulagiri yup! 

@amyrlam and I paired on this today and it _almost_ works. We pulled the `Name` from the `_catalog.json` file, and used a helper to remove the `-##` from the end of the name. This still renders two of each icon, but just in the default size. So need to think through it some more. 

image

MelSumner:
Ok, figured everything out. This PR is ready to review @hashicorp/design-systems

ember-flight-icons/tests/dummy/app/routes/application.js
```diff
@@ -0,0 +1,27 @@
...
+  get contextRootURL() {
+    const config = getOwner(this).resolveRegistration('config:environment');
+    return config.rootURL || '/';
+  }
+
+  async model() {
+    // console.log('here', this.contextRootURL)
+
+    const response = await fetch(
+      `${this.contextRootURL}@hashicorp/ember-flight-icons/icons/_catalog.json`
```
Dhaulagiri, line 15: thought: how will we ensure that this file gets updated?

ember-flight-icons/tests/dummy/app/routes/application.js
```diff
@@ -0,0 +1,27 @@
...
+import { getOwner } from '@ember/application';
+
+export default class ApplicationRoute extends Route {
+  get contextRootURL() {
+    const config = getOwner(this).resolveRegistration('config:environment');
+    return config.rootURL || '/';
+  }
+
+  async model() {
+    // console.log('here', this.contextRootURL)
```
Dhaulagiri, line 12: nitpick (non-blocking): could remove this if it's no longer needed

ember-flight-icons/tests/dummy/app/helpers/substring.js
```diff
@@ -0,0 +1,8 @@
+import Helper from '@ember/component/helper';
+
+export default class Substring extends Helper {
```
Dhaulagiri, line 3: suggestion: add a comment about the intended behavior of this helper

ember-flight-icons/tests/integration/helpers/substring-test.js
```diff
@@ -0,0 +1,17 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Helper | substring', function (hooks) {
+  setupRenderingTest(hooks);
+
+  // TODO: Replace this with your real tests.
+  test('it renders', async function (assert) {
```
Dhaulagiri, line 10: nitpick (non-blocking): if we don't think a real test is necessary we could delete this test file

ember-flight-icons/app/helpers/substring.js
```diff
@@ -0,0 +1 @@
+export { default } from '@hashicorp/ember-flight-icons/helpers/substring';
```
Dhaulagiri, line 1: question: do we want to expose this helper to consumers of the addon?

Dhaulagiri (approved):
👏 

Had a few non-blocking points of feedback/questions to consider before merging

ember-flight-icons/tests/dummy/app/routes/application.js
amyrlam: yeah good one

(non-blocking) want to add, can you in a separate PR consider documenting the https://github.com/hashicorp/flight/tree/main/export-go/dist to https://github.com/hashicorp/flight/tree/main/ember-flight-icons/public/icons workflow, perhaps https://github.com/hashicorp/flight? 

alternatively... i think ideally we just want one source of truth. perhaps we can just use `export-go/dist` as `@hashicorp/flight-icons` ... then have `@hashicorp/flight-icons-ember` (to be renamed, if we agree) consume `@hashicorp/flight-icons`

i'm going to muddle with `@hashicorp/flight-icons` and make another version to npm, but i might try making the file from `export-go/dist` instead today

amyrlam:
👍 looks good, nice one!

separate from this PR, i'd love for us to discuss WIP commits and github stuff, maybe at retro? it was a bit hard to review commit-by-commit. i tend to squash out WIP commits. would love to come to an agreement as a team. 
- in this PR specifically, it would be cool if adding `ember-fetch` and the corresponding package.json / yarn.lock changes as a separate commit so that it's easier to review
- i prob would have dropped the initial commit with all the non-dynamic makeup as well. but that could just be my opinion!
- i've been on teams where everyone uses https://gitmoji.dev/ to keep commits as atomic as possible 

alternatively, we could consider squashing PR's on merge (can be challenging looking back at the history if the PR's are super large, but can always go back to the PR to see the breakdown, no strong opinion)

ember-flight-icons/tests/dummy/app/routes/application.js
Dhaulagiri: right now it's pretty ad hoc (which is ok!) but we need to map out how exactly we want to connect the dots here, and I agree having a source of truth will be better than relying on manual copies. having the file live with the contents of `hashicorp/flight-icons` seems reasonable to me and is similar to what we do with structure-icons currently. For the purposes of documentation generation it _is_ possible to read the contents of the sprite itself and generate the list of icons based off of that, but that's sort of the same difference as reading from that file I suppose.

MelSumner:
@amyrlam I don't have strong opinions on it- I commit and push frequently for...idk, no reason that is super apparent to me now that I think about it. Happy to do whatever makes sense, I don't usually review per commit but rather by overall changes. Totally can adjust.

ember-flight-icons/app/helpers/substring.js
MelSumner: no, they don't need it. we won't (eventually) need it either.

amyrlam:
@MelSumner cool, yeah not trying to derail the PR, just a suggestion, to make it easier / faster for the reviewer. i'll take a stab at some guidelines separately and share so we can start to flesh out our "process" as a team

maybe we can both pair code review in the future, each reviewing something aloud that we haven't previously paired on?

MelSumner:
Sure sounds good!

ember-flight-icons/tests/dummy/app/helpers/substring.js
MelSumner: ok, added 👍

MelSumner:
Update: 

- addressed PR review items
- addressed missing items in the dummy app that were throwing errors in the console
- rebased to be current

Would like to get this merged (squash/merge is fine) and then make other improvements separately.

ember-flight-icons/app/helpers/substring.js
Dhaulagiri: not a blocker because I may be mis-remembering, but isn't there a way to create a helper just for the dummy app?

Dhaulagiri:
:shipit:  I say we land this and followup on any details

ember-flight-icons/tests/integration/helpers/substring-test.js
MelSumner: agree; removed.

Issue #45, opened 2021-08-11
Setup CI for ember-flight-icons


Dhaulagiri:
- [x] Setup using GitHub CI
- [x] Required to pass before changes can be merged
- [ ] Badge added to readme

PR #65, opened 2021-08-17
add ci

Dhaulagiri:
Addresses #45

### Current issues:

- [x] Initial `Lint` run will complete but any subsequent runs that rely on cached data fail -> `/bin/sh: 1: npm-run-all: not found`
- [x] ember try scenarios fail: `node_modules/.bin/ember: No such file or directory`

~~Both seem related to confusion within the job as to where it's running despite setting `working-directory` at the global level~~

Needed to be more clear in specifying `ember-flight-icons` as a path in multiple places in the file, in particular any part that uses `uses` as those do not respect the `working-directory` directive

### Validation

- [x] Initial tests pass (i.e. nothing cached)
- [x] Re-running the job passes (i.e. reading from cache)

### unvalidated
* In theory this will only run on changes to the `ember-flight-icons` path but I am not able to test that locally


MelSumner (approved):
Thank you for setting this up 👍

PR #69, opened 2021-08-17
[not for merge] Test create `flight-icons` package

amyrlam:
Published a 0.0.1-beta here to see if it worked, a lot of this is N/A for non-Ember: https://www.npmjs.com/package/@hashicorp/flight-icons

Part of: https://github.com/hashicorp/flight/issues/60

PR #70, opened 2021-08-17
Update name for npm

amyrlam:
Opening for discussion, definitely not urgent

Made a test `flight-icons` on npm as a 0.0.1-beta for React (is within the Ember directory rn, but just for testing) here https://github.com/hashicorp/flight/pull/69. 

If we go with that name, am wondering if it makes the most sense to rename `ember-flight-icons` to `flight-icons-ember`, "BEM-ish" naming style?  

Second commit is just a find and replace. I can do some spot checking later, just wanted to throw the change idea out there


MelSumner:
Def open to discuss. I wonder if it makes sense to rename it so early, given the other choices we will soon be making. 
It might be a lot of churn to re-name it now only to re-name it something else after that.

amyrlam:
@MelSumner sure sounds good. if i rebase and drop the rename commit, so just https://github.com/hashicorp/flight/pull/70/commits/5d8e98809e583ec327d13d9fe47386faad253cfc, would you be good with that?

for context, i'm interested in publishing a npm beta for ember. i already published the react one, https://www.npmjs.com/package/@hashicorp/flight-icons. if i publish a `0.0.1-beta`, we can only rename it after 72 hours if only one of us publishes it per `has a single owner/maintainer` on https://docs.npmjs.com/policies/unpublish#packages-published-more-than-72-hours-ago

i don't know a lot about unpublishing/renaming packages in npm, but thought that `ember-flight-icons` and `flight-icons-ember` might be confusing

amyrlam:
per the mtg, i'll drop the rename for now and work on https://github.com/hashicorp/design-system-playground-fastboot/issues/6 so folks can test locally

amyrlam:
stale, will open a new branch just correcting the npm file and not the rename for now

PR #73, opened 2021-08-18
added node_modules to gitignore

MelSumner:
If merged, this PR adds the `node_modules` folder to the `.gitignore` file at the root of the repository.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .env
 .DS_Store
+node_modules/


amyrlam:
seems like this doesn't show up in the actions tab https://github.com/hashicorp/flight/actions, not sure why yet

Dhaulagiri:
> seems like this doesn't show up in the actions tab

😖 this is because I set the actions to only run when changes were made to the `ember-flight-icons` path however that seems to be in conflict with the desire to have these checks pass before merging (I assumed that they checks would only appear if they were actually run)

For now any admin (i.e. us) can bypass the checks requirement, but we'll need to figure something out

PR #74, opened 2021-08-18
[not for merge] pairing on fetch scratch

amyrlam:
scratch for iterating thru the catalog.json dynamically but still some changes to make

image

PR #79, opened 2021-08-19
Added search feature to icon display page

### LINKED ISSUE #55: Allow icons index to be searched
Given I can see a list of icons
And I input a search query
I am shown a list of corresponding icons that match that search query

A portion of this functionality will depend on https://github.com/hashicorp/flight/issues/24, so if that is blocking we can initially target exact matches and open a follow-up issue for when that export work has been completed

MelSumner:
## Summary

If merged, this PR resolves #55 by adding the ability to search/filter the icons, as well as some general improvements.

## Details

Changes made:

1. added search/live-filter feature
2. added ability to append css classes to the component when invoked
3. added CSS class info to README
4. added new test for CSS classes
5. tweaked the css so the footer stays at the bottom of the viewport after live-filter results are shown
6. added a few more tests to validate acceptable values for `@color` 
7. added a `onlyname` helper in the dummy app to return the correct name from `catalog.json`

### Screenshots
With search/live-filter implemented:
image

After filter is applied:
image

## Tests
All tests pass


amyrlam:
> Note: I had to change the _catalog.json file in order to make this work reasonably. 

could you elaborate on what you mean by this? did you manually update that file? 

will take a closer look today!

ember-flight-icons/addon/components/flight-icon.hbs
```diff
@@ -1,6 +1,6 @@
 <svg 
   aria-hidden="true"
-  class="flight-icon icon-{{@name}} {{if this.display "display-inline"}}"
+  class="flight-icon icon-{{@name}} {{if this.display "display-inline"}} {{if this.isDemo "demo-icon"}} {{if this.isDemoHidden "d-none"}}"
```
Dhaulagiri, line 3: I would concerned about adding public API/logic like this to the component that is only meant to support the demo app. 

It also seems like this might be an example of something that is going to be common with consumers (wanting to add a class to the svg) which would be good for us to come up with a canonical solution for. What do you think?

MelSumner: @Dhaulagiri I think you are correct. We could position the class attribute such that using `...attributes` would disallow addition of any other attribute but appended class names.

MelSumner:
> could you elaborate on what you mean by this? did you manually update that file?

@amyrlam 

Steps: 
- open the `_catalog.json` file
- find and replace `-16` (with nothing)
- find and replace `-24` (also with nothing)

amyrlam:
@MelSumner thanks for the explanation, it wasn't clear to me what you meant with the file

i think this is a good proof of concept...but not mergeable as-is. to me, a better approach is to work with the file as-is from figma

some ideas:
- you could add instructions to the readme as a part of this PR re: how you manually updated this file
- you could write a script that programmatically updates the file (although we will have to manually do that with every figma export). ideally we want the 700 svg's and the `_catalog.json` to not be duplicated in the repo
- you could explore what the options are with the `_catalog.json` export from the figma side
- maybe some regex solution? we had kinda talked about that in https://github.com/hashicorp/flight/pull/74 for another area of the code

would suggest that you merge in https://github.com/hashicorp/flight/pull/63 (wasn't sure if you were waiting on something re: merging it? there were a few comments from brian but non-blocking)

threw some pairing time on your calendar tmrw if you want to pair

ember-flight-icons/addon/components/flight-icon.hbs
MelSumner: Updated; issue resolved.

MelSumner:
> i think this is a good proof of concept...but not mergeable as-is. to me, a better approach is to work with the file as-is from figma

@amy I agree, which is why I've requested changes to what we get from figma (in the .json) file. 

Totally open to other approaches as well, but this was the one that worked in the cleanest way, IMO. I am up to pair on this early next week; if you wanted to async work on it, a PR to this branch would totally be okay too. 👍

amyrlam (changes requested):
per the convo on the PR, just being clear with "request changes" that this is a WIP and not for merge right now. 

fixing the file from figma sounds good, if you want to explore that approach? i will DM you about pairing next week rn

MelSumner:
> per the convo on the PR, just being clear with "request changes" that this is a WIP and not for merge right now.

@amyrlam yeah sounds good, I'm totally fine with that. I was under the impression that the export was still being figured out, but that the icon display with search was requested sooner rather than later so the dummy app could be deployed and folks could start looking around. If that's not the case, we can leave this PR until the other issues are resolved. Maybe @Dhaulagiri can clarify expectations here? 

FWIW, I planned to run the exporter and update the .json file manually (as part of app/addon maintenance) until the exporting bit was figured out; don't want to block on shipping something because it's not perfect. Would rather know that there are a couple of extra steps and ship; we can always implement iterative improvements. 

Thank you for the feedback 👍

ember-flight-icons/jsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
```
Dhaulagiri, line 1: [personal opinion alert] personal editor config should generally not be committed to a shared repo

ember-flight-icons/tests/dummy/app/helpers/substring.js
```diff
@@ -1,12 +0,0 @@
-import Helper from '@ember/component/helper';
```
Dhaulagiri, line 1: @MelSumner if we keep this helper around does that obviate the need to manually change the catalog file? We could use the substring-ed version of the icon name to do the searching against then, or am I misunderstanding how this worked?

ember-flight-icons/jsconfig.json
MelSumner: The only reason why I committed this file is because it didn’t work without it and the Ember console message said to add it. This made me wonder if we would be able to deploy a working page if this wasn’t in the repo. What do you think?

Dhaulagiri: That sounds 🐟 -y since this is only for development with vs code. I'd be happy to take a look to figure out what might be causing the errors locally.

MelSumner: let me make sure it's not just me, first! 👍

ember-flight-icons/tests/dummy/app/helpers/substring.js
MelSumner: I'll try it and make sure it's not something that I missed 👍

amyrlam:
i am only just starting the vercel task for deployments, so this doesn't seem like a burning fire? if we have vercel deployments, we can view the preview on a branch as well

i'm all for iterative improvements, but it can be a lot of cognitive overload to fix issues once things are merged into main. we already have the go icons copied over to the addon, and they are out of sync now (i manually did that here https://github.com/hashicorp/flight/pull/69) 

i did mention a compromise above:

> you could add instructions to the readme as a part of this PR re: how you manually updated this file

that would be self documenting the hack. but you didn't respond to that, so as-is i don't think this is mergeable. happy to pair on it next week

amyrlam:
made https://github.com/hashicorp/flight/pull/88 so we're mostly not blocked on the icons. didn't get the steps in https://github.com/hashicorp/flight/pull/79#issuecomment-902212675 to work, but ran out of time!

MelSumner:
> i am only just starting the vercel task for deployments, so this doesn't seem like a burning fire? if we have vercel deployments, we can view the preview on a branch as well

Totally. I'm not trying to get this merged exactly this second. Apologies if it came across that way. 

> i'm all for iterative improvements, but it can be a lot of cognitive overload to fix issues once things are merged into main. we already have the go icons copied over to the addon, and they are out of sync now (i manually did that here #69)

I want us to try to have a single source of truth for the icons and import them from that package into their respective addons, like we saw in the `ember-mdi` code. I think it's cleaner and we've all indicated that we want as automated a process as possible. I am still on board to get to that approach. 

> i did mention a compromise above:
> 
> > you could add instructions to the readme as a part of this PR re: how you manually updated this file
> 
> that would be self documenting the hack. but you didn't respond to that, so as-is i don't think this is mergeable. happy to pair on it next week

@amyrlam It's good to compromise when necessary; but I'd rather try to get it working in a way that makes our team happy with it, even if that takes a little bit longer. Plus we made plans to pair and look at it together, hopefully that will help make things more clear.

ember-flight-icons/jsconfig.json
amyrlam: it's unclear to me what the decision is here. it sounds like the decision to remove this file before merge? 

(blocking) please make that change before +1

ember-flight-icons/README.md
```diff
@@ -37,13 +37,13 @@ The `name` must be specified. A default icon size of 24 will be provided.
 
 ### Optional usage
 
-Icons are set to a `viewBox` square size of 24 by default. 
+Icons are set to a `viewBox` square size of 24 by default.
```
amyrlam, line 40: (not blocking) i think we have different editorconfig's here. i've used markdownlint and am pretty happy with it, if we want to go that route

ember-flight-icons/public/icons/_catalog.json
```diff
@@ -1 +1 @@
...
...,"Size":16},{"Name":"watch-24","Size":24},{"Name":"watch-16","Size":16},{"Name":"webhook-24","Size":24},{"Name":"webhook-16","Size":16},{"Name":"wifi-24","Size":24},{"Name":"wifi-16","Size":16},{"Name":"wifi-off-24","Size":24},{"Name":"wifi-off-16","Size":16},{"Name":"wrench-24","Size":24},{"Name":"wrench-16","Size":16},{"Name":"x-24","Size":24},{"Name":"x-16","Size":16},{"Name":"x-circle-24","Size":24},{"Name":"x-circle-16","Size":16},{"Name":"x-circle-fill-24","Size":24},{"Name":"x-circle-fill-16","Size":16},{"Name":"x-diamond-24","Size":24},{"Name":"x-diamond-16","Size":16},{"Name":"x-diamond-fill-24","Size":24},{"Name":"x-diamond-fill-16","Size":16},{"Name":"x-hexagon-24","Size":24},{"Name":"x-hexagon-16","Size":16},{"Name":"x-hexagon-fill-24","Size":24},{"Name":"x-hexagon-fill-16","Size":16},{"Name":"x-square-24","Size":24},{"Name":"x-square-16","Size":16},{"Name":"x-square-fill-24","Size":24},{"Name":"x-square-fill-16","Size":16},{"Name":"zap-24","Size":24},{"Name":"zap-16","Size":16},{"Name":"zap-off-24","Size":24},{"Name":"zap-off-16","Size":16},{"Name":"zoom-in-24","Size":24},{"Name":"zoom-in-16","Size":16},{"Name":"zoom-out-24","Size":24},{"Name":"zoom-out-16","Size":16}]
```
amyrlam, line 1: (blocking) please document this change as requested. i followed your instructions in https://github.com/hashicorp/flight/pull/79#issuecomment-902212675 on the other PR https://github.com/hashicorp/flight/pull/88 and it didn't work for me. i assume you used regex.

amyrlam (changes requested):
i just put time on your calendar for today, if you want to pair. 

the previous blocking comment was not addressed. please address the blocking comments so that the code is maintainable. i don't really like "requesting changes", but i'm making that review to be clear

ember-flight-icons/public/icons/_catalog.json
MelSumner: @amyrlam This is the file straight from the export, I replaced the one where I'd made changes. So this is the raw file with no changes.

ember-flight-icons/README.md
MelSumner: I have markdownlint as well. Happy to sync settings, I just use the default settings.

amyrlam: https://github.com/hashicorp/flight/issues/92

ember-flight-icons/public/icons/_catalog.json
amyrlam: gotcha, thanks for the explanation, that wasn't clear to me as a reviewer that you made that change

amyrlam (approved):
lgtm

amyrlam:
not sure about the failing test, thanks for making the changes

MelSumner:
> not sure about the failing test, thanks for making the changes

I've confirmed with others that the ember-canary failures are on the Ember side and not just us; will need to look into what Embroider wants but I think it is low-risk to move forward since we're not embroider optimizing right now anyway. 👍

amyrlam:
documented that here so it's clear: https://github.com/hashicorp/flight/pull/95

PR #82, opened 2021-08-20
Update the package.json

amyrlam:
Extracted from: https://github.com/hashicorp/flight/pull/70

Made this tweak based on me publishing @hashicorp/flight-icons in https://github.com/hashicorp/flight/pull/69.


MelSumner (approved):
Thank you for doing this!

PR #86, opened 2021-08-20
Internal: Add pull request template

MelSumner:
## Summary

If merged, this PR adds a pull request template per [GitHub's instructions](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates) and some sensible defaults. 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
 .env
 .DS_Store
-node_modules
+node_modules/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
 .env
 .DS_Store
-node_modules/
+node_modules
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,0 +1,13 @@
+## Summary
+
+<!-- Required section. Describe what this PR will do if it is merged. -->
+
+## Details
+
+<!-- Optional section. Give more details if it is sensible to do so. -->
+
+## Tests
+
+<!-- Required section. Run the tests and ensure that they pass. If only specific tests were run for this PR, indicate that here. -->
+
+Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -17,10 +17,6 @@ Examples:
 
 <!-- If more details are appropriate, add them here. What code changed, and why? -->
 
-
-
-
-
 ## :camera_flash: Screenshots
 
 <!-- Screenshots always help, especially if this PR will change what renders to the browser -->
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -18,7 +18,6 @@ Examples:
 <!-- If more details are appropriate, add them here. What code changed, and why? -->
 
 
-## :flags: Feature Flag
 
 <!-- Include any related feature flag names if appropriate. -->
 
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,13 +1,45 @@
-## Summary
+## :pushpin: Summary
 
-<!-- Required section. Describe what this PR will do if it is merged. -->
-
-## Details
-
-<!-- Optional section. Give more details if it is sensible to do so. -->
-
-## Tests
-
-<!-- Required section. Run the tests and ensure that they pass. If only specific tests were run for this PR, indicate that here. -->
-
+<!-- If merged, this PR.... 
+This should be a short TL;DR that includes the purpose of the PR.
+-->
+### :speech_balloon: 
 Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.
+
+<!--
+Examples: 
+- issue (ux,non-blocking): These buttons should be red, but let's handle this in a follow-up.
+- suggestion (non-blocking): Let's change this wording to make it easier to understand.
+- issue (blocking): We shouldn't introduce this kind of tech debt; let's pair and resolve the issue in a more sustainable way.
+-->
+
+## :hammer_and_wrench: Detailed Description
+
+<!-- If more details are appropriate, add them here. What code changed, and why? -->
+
+
+## :flags: Feature Flag
+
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -1,3 +1,3 @@
 .env
 .DS_Store
-node_modules/
+node_modules
```
amyrlam, line 3:
```suggestion
node_modules/
```

i think it should include the slash re: https://stackoverflow.com/questions/29820791/git-ignore-node-modules-folder-everywhere and https://github.com/github/gitignore/blob/master/Node.gitignore#L42, is that right or were you experiencing something different?

to overcommunicate, this change doesn't seem relevant to the PR title. isn't mentioned in the commit message. would be easier to review as a separate commit or PR, if is an intended change. definitely no big deal but to communicate what's going through my mind when reviewing the PR

pull_request_template.md
```diff
@@ -0,0 +1,13 @@
+## Summary
+
+<!-- Required section. Describe what this PR will do if it is merged. -->
+
+## Details
```
amyrlam, line 5: "Details" seems maybe redundant to "Summary" and the PR title? 

what do you think about the sample here? https://eng-handbook.hashicorp.services/git-and-github/#templates

pull_request_template.md
```diff
@@ -0,0 +1,13 @@
...
+
+<!-- Optional section. Give more details if it is sensible to do so. -->
+
+## Tests
```
amyrlam, line 9: not sure what you are expecting folks to write in "Tests", could you elaborate? seems like is handled as CI?

pull_request_template.md
```diff
@@ -0,0 +1,13 @@
...
+
+<!-- Required section. Run the tests and ensure that they pass. If only specific tests were run for this PR, indicate that here. -->
+
+Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.
```
amyrlam, line 13: seems useful, but would be nice to have inline so it's a better DX vs TLDR.

i like what cloud-ui is doing, and seems similar to conventional comments. is very clear to any developer what to do as it's written out in the template. what is your opinion on it? https://github.com/hashicorp/cloud-ui/blob/master/.github/pull_request_template.md

amyrlam (commented):
left a few comments for discussion! 

i think the .gitignore should be fixed before merge, lmk if i'm wrong there

pull_request_template.md
MelSumner: I don't agree that they are the same thing. Often a summary should suffice; however there are times when a PR requires more detailed information be added for context. That's why this section is optional and can be removed.

amyrlam: sure sounds good! i typically push that summary/detail hierarchy to the commit message. i often just copy my commit messages to the PR summary

here's an example of what i mean from https://eng-handbook.hashicorp.services/git-and-github/#commit-messages 

```
Example commit message:

feature: Allow users to navigate to X from the sidebar

This feature was necessary since a customer wanted to be able to quickly access X while in the UI. We will probably want to break out this feature into the component library so other products can use it as well.
```

pull_request_template.md
MelSumner: That’s nice! I don’t like the mountain/pebble/dust thing though. More things I have to translate the relative meanings of- I think it’s more useful to use the exact word instead. We could otherwise simply adopt their PR template, it seems fine. Would that work for you?

amyrlam: sure. what about taking their template but adding some inline documentation to conventional comments? e.g. 

```suggestion
Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR. For example:

issue (blocking): 
nitpick (non-blocking):
```

pull_request_template.md
```diff
@@ -0,0 +1,45 @@
...
+- suggestion (non-blocking): Let's change this wording to make it easier to understand.
+- issue (blocking): We shouldn't introduce this kind of tech debt; let's pair and resolve the issue in a more sustainable way.
+-->
+
+## :hammer_and_wrench: Detailed Description
+
+<!-- If more details are appropriate, add them here. What code changed, and why? -->
+
+
+## :flags: Feature Flag
```
amyrlam, line 21:
```suggestion
```

we don't have feature flags rn. will suggest we remove this until we do?

amyrlam (approved):
👍 left a non-blocking suggestion

pull_request_template.md
```diff
@@ -0,0 +1,44 @@
...
+- issue (blocking): We shouldn't introduce this kind of tech debt; let's pair and resolve the issue in a more sustainable way.
+-->
+
+## :hammer_and_wrench: Detailed Description
+
+<!-- If more details are appropriate, add them here. What code changed, and why? -->
+
+
+
+<!-- Include any related feature flag names if appropriate. -->
```
MelSumner, line 22:
```suggestion

```

pull_request_template.md
```diff
@@ -0,0 +1,43 @@
...
+-->
+
+## :hammer_and_wrench: Detailed Description
+
+<!-- If more details are appropriate, add them here. What code changed, and why? -->
+
+
+
+
+
```
MelSumner, line 23:
```suggestion
```

pull_request_template.md
```diff
@@ -0,0 +1,39 @@
+## :pushpin: Summary
+
+<!-- If merged, this PR.... 
+This should be a short TL;DR that includes the purpose of the PR.
+-->
+### :speech_balloon: 
+Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.
+
+<!--
+Examples:
```
Dhaulagiri, line 10: opening my first PR with this and having these be comments doesn't seem helpful since they are for the reviewer, not the person who is opening the PR. The person reviewing the PR won't see these. Should we move these into the actual text of the PR template? @MelSumner wdyt?

PR #94, opened 2021-08-23
Changes default icon size to 16x16

### LINKED ISSUE #78: Default icon size should be 16
There's been some miscommunication around this, but the default icons size should be 16 and not 24. We'll need to update the component and the documentation to reflect this.

MelSumner:
## :pushpin: Summary

If merged, this PR resolves #78 

This PR changes the default size of the icon from 24x24 to 16x16. The tests and documentation were also updated to reflect the changes made in the component.

Note: Two test scenarios, `ember-canary` and `embroider-optimized` are not currently passing. We have confirmed that the `ember-canary` scenario is widespread and not due to this addon. We are looking into the `embroider-optimized` scenario but as we are not currently using embroider or optimizing for it, it is not a blocker if this scenario does not pass right now. 

## :+1: Definition of Done

- [x] New/updated functionality works 
- [x] Tests added/updated
- [x] Docs updated

### DIFF
--- a/ember-flight-icons/README.md
+++ b/ember-flight-icons/README.md
@@ -29,7 +29,7 @@ ember install @hashicorp/ember-flight-icons
 
 ### Required usage
 
-The `name` must be specified. A default icon size of 24 will be provided.
+The `name` must be specified. A default icon size of 16 is provided.
 
 ```hbs
 <FlightIcon @name="activity" />
--- a/ember-flight-icons/README.md
+++ b/ember-flight-icons/README.md
@@ -43,7 +43,7 @@ Icons are set to a `viewBox` square size of 16 by default.
 <FlightIcon @name="activity" />
 ```
 
-Currently, we support a `viewBox` square size of 24 or 16. If you need your icon to be larger than 16, pass in `size`.
+Currently, we support a `viewBox` square size of 16 or 24. If you need your icon to be larger than 16, pass in `size`.
 
 ```hbs
 <FlightIcon @name="activity" @size="24" />
--- a/ember-flight-icons/README.md
+++ b/ember-flight-icons/README.md
@@ -37,16 +37,16 @@ The `name` must be specified. A default icon size of 24 will be provided.
 
 ### Optional usage
 
-Icons are set to a `viewBox` square size of 24 by default.
+Icons are set to a `viewBox` square size of 16 by default.
 
 ```hbs
 <FlightIcon @name="activity" />
 ```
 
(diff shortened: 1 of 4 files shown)


ember-flight-icons/README.md
```diff
@@ -37,16 +37,16 @@ The `name` must be specified. A default icon size of 24 will be provided.
 
 ### Optional usage
 
-Icons are set to a `viewBox` square size of 24 by default.
```
Dhaulagiri, line 40: there's a reference to default of 24px up on line 32 ☝️  that we should update to 16

ember-flight-icons/README.md
```diff
@@ -37,16 +37,16 @@ The `name` must be specified. A default icon size of 24 will be provided.
...
+Icons are set to a `viewBox` square size of 16 by default.
 
 ```hbs
 <FlightIcon @name="activity" />
 ```
 
-Currently, we support a `viewBox` square size of 24 or 16. If you need your icon to be smaller than 24, pass in `size`.
+Currently, we support a `viewBox` square size of 24 or 16. If you need your icon to be larger than 16, pass in `size`.
```
amyrlam, line 46:
```suggestion
Currently, we support a `viewBox` square size of 16 or 24. If you need your icon to be larger than 16, pass in `size`.
```

suggestion: for clarity list default size first

ember-flight-icons/README.md
didoo: Am I wrong, or do I remember yesterday we discussed this and we agreed to make the size mandatory (without a default) and not have a predefined size at code level, only in Figma?

MelSumner: @didoo I remember discussing it as an option, but I didn't think we'd made a hard decision on it. So far it seems like we're trying to match the current developer experience, which seems to be to provide strong defaults with customizable options for flexibility when needed. I wonder if changing that approach would mean a change via RFC? What do you think?

amyrlam: yesterday, i thought we agreed to change size to 16. should we consider the size being required in a separate issue? getting this in for the smoke testing folks are doing in consuming apps rn seems prudent?

amyrlam (approved):
👍 lgtm, but in case we want to discuss at the meeting to move forward, will let you merge

amyrlam:
we are ignoring the failed `embroider-optimized` test (is allowed to fail on main)

per the meeting, we'll merge this for now and could address a required size on this issue https://github.com/hashicorp/design-systems-team/issues/24

PR #95, opened 2021-08-23
Update template to mention why failing tests

amyrlam:
## :pushpin: Summary

Provide a better DX

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -30,6 +30,8 @@ Examples:
 
 <!-- List steps to test your change on a local environment. -->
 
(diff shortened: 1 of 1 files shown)


Dhaulagiri:
This seems expedient, but I also think it might be worth evaluating whether or not we actually need to test against canary or the embroider setup given we are unlikely to want to invest time into fixing a failure that only happens against canary for example.

MelSumner (approved):
suggestion (non-blocking): I wouldn't mind it if we took those test scenarios out of our ember-try setup.

amyrlam:
seems like could merge in https://github.com/hashicorp/flight/pull/96, instead of this PR

amyrlam:
closing in favor of https://github.com/hashicorp/flight/pull/96

PR #96, opened 2021-08-23
allow canary and embroiderOptimized scenarios to fail

Dhaulagiri:
## :pushpin: Summary

Addresses https://github.com/hashicorp/flight/pull/95 by allowing the flaky tests to fail without failing the overall build

Screen Shot 2021-08-23 at 15 47 44

## :+1: Definition of Done
- [x] The scenarios in question fail in their runs but don't fail the overall action
- [x] We are happy with this approach vs just deleting the tests entirely

### DIFF
--- a/ember-flight-icons/config/ember-try.js
+++ b/ember-flight-icons/config/ember-try.js
@@ -41,6 +41,7 @@ module.exports = async function () {
       },
       {
         name: 'ember-canary',
+        allowedToFail: true,
         npm: {
           devDependencies: {
             'ember-source': await getChannelURL('canary'),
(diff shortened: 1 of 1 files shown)


amyrlam (approved):
👍 sgtm 

just noting for posterity, i see your screenshot here https://github.com/hashicorp/flight/runs/3404709744?check_suite_focus=true#step:7:175

MelSumner (approved):
I like this approach the most of the ones we've considered. This gives us a way to keep an eye on it and perhaps even make improvements that will make those scenarios pass.

Dhaulagiri:
Will merge but one thing to keep in mind is that because these are failing green you actually have to inspect them to see if the tests are working or not which limits their usefulness. Something for future us to revisit.

PR #101, opened 2021-08-24
Docs design tweaks

dizzyup:
## :pushpin: Summary

Adds the latest version of the exported icons (#100) to the dummy app/docs and updates the page styles a little.

- Copies the updated `/dist`assets from #100 to the dummy app
- Aligned the styles to the Figma file (still intended to be temporary)

## :camera_flash: Screenshots

CleanShot 2021-08-24 at 15 21 02@2x


ember-flight-icons/tests/dummy/app/templates/application.hbs
```diff
@@ -35,12 +33,15 @@
...
-            @name={{Onlyname meta.name}}
-            @size={{if (eq meta.size 16) '16'}}
-            class="demo-icon"
-          />
-          {{meta.name}}
+          <div class="icon-frame">
+            <FlightIcon
+              @name={{Onlyname meta.name}}
+              @size={{if (eq meta.size '16') '16'}}
+              @color="currentColor"
```
dizzyup, line 40: @MelSumner I think something in the icon component logic might be off - I couldn't inherit colours unless I explicitly passed `currentColor` to the `@color` property which is why I added this line.

dizzyup: Looking at [this line](https://github.com/hashicorp/flight/blob/c0ac030f259bdf11c4c761a76fa83c584a73e8ee/ember-flight-icons/addon/components/flight-icon.js#L19) in the component logic it seems like it should be falling back to `currentColor` but I wasn't seeing that in the dummy app when using the component - it didn't seem to output a `fill` property on the SVG DOM element unless explicitly defined.

ember-flight-icons/tests/dummy/app/styles/app.css
```diff
@@ -60,46 +65,67 @@ h2.ds-h2 {
...
+  border-radius: 3px;
+  box-shadow: 0 0 0 1px var(--gray-5);
 }
 ul.ds-ul-grid li.ds-li:hover {
-  background-color: aliceblue;
+  background-color: var(--info-l2);
+  box-shadow: 0 0 0 1px var(--info-l1);
 }
-ul.ds-ul-grid li.ds-li svg.flight-icon {
+ul.ds-ul-grid li.ds-li .icon-frame {
```
MelSumner, line 99: @dizzyup I have been trying to prefix all classes in the demo app with `.ds-` to get into the practice, that way we avoid naming collisions with proper components in the future.

dizzyup: 👍 Fixed!

I was going to look at moving it to tailwind but thought better of it and decided to save that bigger discussion for another day 😉

ember-flight-icons/tests/dummy/app/templates/application.hbs
MelSumner: oh I see what it is... in `flight-icon.hbs` I think it should be `fill={{this.color}}` - right now, the template says "only give me what is passed to me". I'll fix that in a separate PR. 👍

MelSumner: ok: https://github.com/hashicorp/flight/pull/103

amyrlam (approved):
i rebased the PR

👍 lgtm  

what are you thinking @MelSumner, are you good with the PR? am interested in getting the PR in, so that we could publish a 0.0.1-beta `@hashicorp/ember-flight-icons` before the guild meeting tomorrow, if there is time ofc

ember-flight-icons/tests/dummy/app/templates/application.hbs
MelSumner:
```suggestion
```
Can you try removing this line, you shouldn't need it after the bug fix. 👍

MelSumner (changes requested):
Just one small change, thank you!

MelSumner (approved):

PR #103, opened 2021-08-24
bug: resolved issue where currentColor was not being set properly

MelSumner:
## :pushpin: Summary

If merged, this PR resolves the issue where currentColor was not being inherited correctly.

- updated component template to pass `currentColor` correctly
- added tests to check that the default works as expected, and that currentColor is inherited correctly.

### DIFF
--- a/ember-flight-icons/addon/components/flight-icon.hbs
+++ b/ember-flight-icons/addon/components/flight-icon.hbs
@@ -3,7 +3,7 @@
   ...attributes
   aria-hidden="true"
   data-test-icon
-  fill={{@color}} 
+  fill="{{this.color}}"
   height="{{this.size}}"
   id={{this.iconId}}
   viewBox="0 0 {{this.size}} {{this.size}}"
--- a/ember-flight-icons/tests/integration/components/flight-icon-test.js
+++ b/ember-flight-icons/tests/integration/components/flight-icon-test.js
@@ -60,6 +60,21 @@ module('Integration | Component | flight-icon', function (hooks) {
       fill: 'rgb(186, 34, 38)',
     });
   });
-  // TODO: the component should have `color` set to `currentColor` by default
+  // the component should have `color` set to `currentColor` (black) by default
+  test('the fill should be set to black by default', async function (assert) {
+    await render(hbs`<FlightIcon @name="meh" />`);
+    assert.dom(`svg.flight-icon`).hasStyle({
+      fill: 'rgb(0, 0, 0)',
+    });
+  });
+  // currentColor should change per inheritance rules
+  test('The fill color should be able to be inherited from parent', async function (assert) {
+    await render(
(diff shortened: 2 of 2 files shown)


ember-flight-icons/addon/components/flight-icon.hbs
```diff
@@ -3,7 +3,7 @@
   ...attributes
   aria-hidden="true"
   data-test-icon
-  fill={{@color}} 
+  fill="{{this.color}}"
```
amyrlam, line 6: may have been left over from me, my bad!

ember-flight-icons/tests/integration/components/flight-icon-test.js
```diff
@@ -60,6 +60,21 @@ module('Integration | Component | flight-icon', function (hooks) {
...
-  // TODO: the component should have `color` set to `currentColor` by default
+  // the component should have `color` set to `currentColor` (black) by default
+  test('the fill should be set to black by default', async function (assert) {
+    await render(hbs`<FlightIcon @name="meh" />`);
+    assert.dom(`svg.flight-icon`).hasStyle({
+      fill: 'rgb(0, 0, 0)',
+    });
+  });
+  // currentColor should change per inheritance rules
+  test('The fill color should be able to be inherited from parent', async function (assert) {
```
amyrlam, line 71: thanks for making the tests, this is awesome!

suggestion: consider having the comment and test title be one, for code maintainability and clarity

```suggestion
  test('the component should have `color` set to `currentColor` (black) by default', async function (assert) {
    await render(hbs`<FlightIcon @name="meh" />`);
    assert.dom(`svg.flight-icon`).hasStyle({
      fill: 'rgb(0, 0, 0)',
    });
  });
  test('The fill color (currentColor) should be able to be inherited from parent', async function (assert) {
```

amyrlam (approved):
👍 lgtm

i'm not sure if the default color should be black (would think grey, but not sure which one), but seems like a detail that could be addressed separately, if is an issue? mentioning for conversation

MelSumner:
> i'm not sure if the default color should be black (would think grey, but not sure which one), but seems like a detail that could be addressed separately, if is an issue? mentioning for conversation

Black is the default color in the browser if no color is defined. I suppose that's more specifically what I meant by default. In our case, we default to `currentColor`, whatever that happens to be. For the purposes of the test, then, it would be black. What do you think?

amyrlam:
@MelSumner sgtm, that makes total sense

ember-flight-icons/tests/integration/components/flight-icon-test.js
MelSumner: @amyrlam do you mind if I put that in a separate PR? I have comments above all of the tests, not just these ones. Happy to remove all of them at once.

amyrlam: @MelSumner yes sure, sgtm, thanks!

Issue #26, opened 2021-08-09
Document contribution guidelines


Dhaulagiri:
_more meta about how to contribute to the project vs contributing specific icon set changes_

PR #105, opened 2021-08-24
Update docs

amyrlam:
## :pushpin: Summary

Document the manual processes we are currently using

I'm making this PR because I'm going to propose a 0.0.4-beta @hashicorp/flight-icons npm package in a separate PR (haven't been committing the npm tarbell file for that work yet, wasn't the best practice.


MelSumner (approved):
[question, non-blocking] Do you think we should indicate that consumers should not have to do these steps? Or perhaps that these are instructions for the addon maintainers? Just a thought.

amyrlam:
@MelSumner let's get the docs up as-is since not sure where the package.json lives for @hashicorp/flight-icons yet. i noted your comment on https://github.com/hashicorp/flight/issues/26 so that it's tracked. maybe the README links to a CONTRIBUTING file.

PR #104, opened 2021-08-24
Automate reviewers

amyrlam:
## :pushpin: Summary

* Move PR template into .github
  * Colocate GitHub files re: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository
* Add codeowners file
  * Will automate assigning the team as reviewers to any PR. re: https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-code-owners

## :link: External Links

Part of: hashicorp/design-systems-team#23

PR #106, opened 2021-08-24
Fix CODEOWNERS functionality

amyrlam:
## :pushpin: Summary

I missed the `*` before.

Fix to: https://github.com/hashicorp/flight/pull/104
Ref: https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-code-owners


MelSumner (approved):
Went back through the documentation for this to make sure we didn't miss anything else- looks like we can really get some granularity in there, which I think might be useful in the future. very cool! 👍

PR #107, opened 2021-08-24
Document the current (manual) npm process

amyrlam:
## :pushpin: Summary

if merged, we should run a `npm publish`. i previously hadn't committed the tarbell files for the previous versions to `0.0.4-beta` (still not published) 

there's a lot of manual steps here right now, but you can follow commit-by-commit

- we should definitely automate the processes (scripts in the package.json?), but this isn't documented at all rn
- i'm not sure if the `@hashicorp/flight-icons` package.json should live in the root, thoughts? 
- relates to some of the README updates in https://github.com/hashicorp/flight/pull/105/files
- this PR builds off of https://github.com/hashicorp/flight/pull/101


amyrlam:
let's get https://github.com/hashicorp/flight/pull/101 in then rebase this one

amyrlam:
@Dhaulagiri this needs to be rebased, will wait for the other PR 

this is what i was talking about re: where the package.json lives. i think you are suggesting making a `flight-icons/` dir at root level that contains the package.json, npm tarball, and a README, is that right? 

AFAICT you generate the tarball from ember-flight-icons/public/icons rn. that's where i've put the generate the sprite docs too (for now)

i made this issue based on what i tagged you and @MelSumner in yesterday https://github.com/hashicorp/flight/issues/108. i def want to do that, but i don't know how to achieve it yet technically

PR #109, opened 2021-08-24
Tweak to the PR template

amyrlam:
## :pushpin: Summary

i think we want the examples to appear so visible to the reviewer, what do you think


MelSumner (approved):
Sure, seems good! 👍

PR #119, opened 2021-08-28
[🚧 WIP] Update usage instructions for -16 vs -24 icons

amyrlam:
## :pushpin: Summary

First stab at instructions based on https://github.com/hashicorp/boundary-ui/pull/720 discussion on PR, Slack, and in person. Looking for feedback!

Review commit-by-commit

Examples: 

## 👉 Next steps

- [table for now] TODO: File ticket if seems like a good incremental next step? Ideally we could embed flight-icons/README and ember-flight-icons/README to display on https://flight-staging.vercel.app/. We can also update all 3 in tandem, in case we want to display stuff differently on https://flight-staging.vercel.app/ ... if as a team we are cool with that?

### DIFF
--- a/flight-icons/README.md
+++ b/flight-icons/README.md
@@ -10,15 +10,17 @@ yarn add @hashicorp/flight-icons
 
 ## Usage
 
-### How to choose 16 or 24 viewBox size
+We intend for icons to be used at a 16 or 24 `viewBox` size.
 
-* If an icon is in a viewBox <= 16, use the `-16` version of the icon.
-* If an icon is in a 16 < viewBox < 24, use your discretion. <!-- TBH I'm not sure, just put in a placeholder. -->
-* If an icon is in a viewBox >= 24, use the `-24` version of the icon.
+### How to choose 16 or 24 `viewBox` size
+
+* If an icon is in a `viewBox` <= 16, use the `-16` version of the icon.
+* If an icon is in a 16 < `viewBox` < 24, default to the `-16` version of the icon. If you're not sure, please ping the Design Systems team: [@hashicorp/design-systems on GitHub](https://github.com/orgs/hashicorp/teams/design-systems) or [#team-design-systems on Slack](https://app.slack.com/client/T024UT03C/C7KTUHNUS/thread/C7KTUHNUS-1630354922.058700).
+* If an icon is in a `viewBox` >= 24, use the `-24` version of the icon.
 
 ### Guidelines for other sizes
 
-* If you're scaling an icon with a parent component or styling, do not stretch a `-16` icon to a `24` viewBox.
+* If you're scaling an icon with a parent component or styling, do **not** stretch a `-16` icon to a `24` viewBox.
 
 ## Contributing
 
--- a/flight-icons/README.md
+++ b/flight-icons/README.md
@@ -1,6 +1,6 @@
 # flight-icons
 
-An addon for the icons from `flight`. Currently in development.
+An addon for the icons from `flight`. Currently in development. This addon can be used in React or Ember apps.
 
 ## Installation
 
--- a/flight-icons/README.md
+++ b/flight-icons/README.md
@@ -1,0 +1,33 @@
+# flight-icons
+
+An addon for the icons from `flight`. Currently in development.
+
+## Installation
+
+```bash
+yarn add @hashicorp/flight-icons
+```
+
+## Usage
+
+### How to choose 16 or 24 viewBox size
+
+* If an icon is in a viewBox <= 16, use the `-16` version of the icon.
+* If an icon is in a 16 < viewBox < 24, use your discretion. <!-- TBH I'm not sure, just put in a placeholder. -->
+* If an icon is in a viewBox >= 24, use the `-24` version of the icon.
+
+### Guidelines for other sizes
+
+* If you're scaling an icon with a parent component or styling, do not stretch a `-16` icon to a `24` viewBox.
+
+## Contributing
+
+🚧 TBD
+
+## License
+
+This project is licensed under the [Mozilla Public License 2.0](LICENSE.md).
+
+## Versioning
+
+We use [SemVer](http://semver.org/) for versioning.
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Please see [ember-flight-icons/README](ember-flight-icons/README.md).
 
 🚨 Note: npm addon is currently in beta and not intended for production use yet.
 
-This addon can be used in React or Ember apps.
+Please see [flight-icons/README](flight-icons/README.md).
 
 ## Development instructions for `@hashicorp/ember-flight-icons` and `@hashicorp/flight-icons`
 


flight-icons/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Installation
+
+```bash
+yarn add @hashicorp/flight-icons
+```
+
+## Usage
+
+### How to choose 16 or 24 viewBox size
```
MelSumner, line 13: I am not sure this section is accurate. I think the viewbox always stays the same but the height and width can change. See [this CodePen](https://codepen.io/melsumner/pen/1143304b4933a757d0a1e8233881bba7?editors=1100) and also [this PR](https://github.com/hashicorp/flight/pull/126).

MelSumner (changes requested):
I think we need to put this on hold until tomorrow, and let Hector review the sizing issue. Looks good otherwise. 👍

flight-icons/README.md
amyrlam: yeap - not sure why you did request changes, this is a draft pr so just looking for discussion. to start it off, i made the pr to collect feedback from the team

amyrlam: added this PR to the sync notes for tomorrow. not crosslinking here cause internal link

flight-icons/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+```bash
+yarn add @hashicorp/flight-icons
+```
+
+## Usage
+
+We intend for icons to be used at a 16 or 24 `viewBox` size.
+
+### How to choose 16 or 24 `viewBox` size
```
amyrlam, line 15: @dizzyup @didoo took a stab at this from the meeting, what do you think of this wording as a first stab? also anyone is open to reply, but thought you were the most relevant

currently we have no readme for https://www.npmjs.com/package/@hashicorp/flight-icons, so was thinking some initial guidance would be better than nothing, so that we can empower external teams to test the addons on their own

flight-icons/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Installation
+
+```bash
+yarn add @hashicorp/flight-icons
+```
+
+## Usage
+
+We intend for icons to be used at a 16 or 24 `viewBox` size.
```
didoo, line 13: suggestion: maybe we can make the text more clear

Maybe I am missing something, but from what I understand here we're saying/assuming that the only way to consume the icons is via <SVG><use/></SVG>? otherwise the word "viewBox" used in this way can be misleading (it is/was for me, when I read it the first time). Could we use a more generic word, like container box, or parent container, or something similar?

In this context ("Usage") maybe we could be even just say:
_We intend for icons to be used at a 16px or 24px size_
(this is what @dizzyup intended, from what I understand) and then explain the sizing logic (below) in a more general way, about the sizes as expressed in the design/UI and which of the two variants (16 vs 24) use depending on the case.

flight-icons/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+### How to choose 16 or 24 `viewBox` size
+
+* If an icon is in a `viewBox` <= 16, use the `-16` version of the icon.
```
didoo, line 17: issue: while reading, I am not sure what to do with this information, is not clear for me (as a developer)

I am not 100% sure about what we want to say here (also in relation to the second part below), but a few thoughts while reading it:
- the word "viewBox" reminds me of the SVG "viewBox" property, but from what I understand the icons can be used in different ways, right? e.g. in _react-components_ icons are used via `<InlineSvg src={svgFile} />` so in that case I don't think the size is set via the viewBox, right?
- here we say that for intermediate sizes (16 < x < 24) always use 16, but below we have a different criteria; I think the two are kinda antithetic? I suspect we should give the same advice (probably even merge them, as suggested below)

didoo (commented):
@amyrlam added a few comments, if you want we can talk more in details tomorrow, after @dizzyup replies

flight-icons/README.md
```diff
@@ -0,0 +1,35 @@
...
+* If an icon is in a 16 < `viewBox` < 24, default to the `-16` version of the icon. If you're not sure, please ping the Design Systems team: [@hashicorp/design-systems on GitHub](https://github.com/orgs/hashicorp/teams/design-systems) or [#team-design-systems on Slack](https://app.slack.com/client/T024UT03C/C7KTUHNUS/thread/C7KTUHNUS-1630354922.058700).
+* If an icon is in a `viewBox` >= 24, use the `-24` version of the icon.
+
+### Guidelines for other sizes
+
+* If you're scaling an icon with a parent component or styling, do **not** stretch a `-16` icon to a `24` viewBox.
```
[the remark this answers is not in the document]
amyrlam: yeah... i've been confused cause the height and width don't have units. so i've been using the `viewBox` as the source of truth. FWIW, definitely happy for all of this verbiage to be trashed, just wanted to get the ball rolling based on questions that have been coming up

amyrlam:
just wanted to add, opening this more for guidance, feel free to commit directly to this branch and edit at will / or make your own

it was already super useful talking about it at the meeting today, thanks for chiming in!

didoo:
@amyrlam OK then, I'll take a stab at it and see what I can come up with, then ask you (and @dizzyup ) a second opinion.

johncowen:
Hey 👋 not sure if we are looking for further observations here from folks using this? So feel free to ignore if it's not something folks are after yet (and lemme know so I don't keep commenting!). I was gonna add some inlines, but as I wrote it made more sense to add on the end here.

Thanks for putting this documentation together, it really helps to communicate how we might consume these icons, it's good to see folks working on this.

My take on this is:

From what I remember from the RFC (I remember specifically asking a comment about this), the icons should never be scaled.

To me "If you are scaling an icon" suggests we are fine with folks scaling them. "don't do this" is clearer for me as a developer, not only that but its clear to me that this is an important thing from the formatting used here, and I think thats the message I got from the RFC and the README here. I guess @dizzyup can clarify whether scaling is ok or not.

SVGs and `px` are always confusing unfortunately. Just to give some background. We went through a "width x height" vs "viewBox" thing a while back, and the decisions around there where lost in space/time. The initial experiment I did with flight and Consul "regressed" (if you can call it that) to the width x height thing, so we immediately had a bunch of cut off icons when we upgraded. We've done a little automated tweaking to bring that back and reported that to theDS  team. I've think I've seen problems in other products lately too regarding no viewBox.

For me as a developer, naming is, well naming. I'm personally easy with either, I'll probably talk about them as `16` and `24` whichever way you folks go here.

Lastly, the:

>  icons are used via <InlineSvg src={svgFile} /> so in that case I don't think the size is set via the viewBox, right?

Unless I misunderstood this package only contains SVGs? So no frameworks at all, no react, no ember, no nothing just SVGs. Looking at the published npm package the width and heights are specified but no viewBox. Consul-wise we are fine either way here (we can just automate out/in the width/height vs viewBox). But thought I'd mention as I'm not sure if that is intentional or not?

Oh, one last thing (sorry!) - I also noticed that the published package doesn't seem to contain `svgo`ed SVGs? I think I saw somebody say somewhere that they would be pre-compressed?

Again, please let me know if this is the kind of feedback you are looking for or not, ty!

amyrlam:
@johncowen hey, feel free to chime in whenever i would say! 

icons should not be scaled in new code. but... (1) there may be exceptions in new code. (2) that may not be possible / feasible in all legacy code.

we've run into situations already where the icon is in line with typography that's 20px high in `boundary-ui`, or a smaller icon like 11px is being used in `learn`. so... looking for us to flesh out the guidelines as we're trying to build adoption of the package

there are 2 packages: https://www.npmjs.com/package/@hashicorp/ember-flight-icons and https://www.npmjs.com/package/@hashicorp/flight-icons. i _think_ you may be able to use flight-icons with the css approach in consul, but i haven't dug into it yet, we could pair next week if you want? (i'll be out fri & tue, extending the long weekend)

i thought the icons were svgo'd, can you be more specific about what are you seeing? i can file that as a separate bug ticket if can get more detail! also, feel free to file atomic issues on `flight` at will, helps for tracking

junyper:
FWIW, I still think the icons should allow for scaling for a11y reasons, and using `sm` and `med` vs `16` and `24` in the naming would allow for changing the dimensions later on w/out having to update all consuming apps.

amyrlam:
@junyper thanks for the feedback!
- we went with the #s for now, cause naming things is hard. i do like how tailwind uses sm/md/lg...but it's a more fleshed out system vs what we have right now. renaming #s to names is something that could potentially happen down the road
- there was a spike on https://github.com/hashicorp/flight/issues/30 but it was closed, i think we'll keep it in the icebox for now
- if you have examples in existing code on `atlas` with scaling, we can definitely look at them. we sort of need to define what are guidelines are for new designs vs. existing designs. i started this PR as these questions came up from `boundary` and `learn` using https://www.npmjs.com/package/@hashicorp/flight-icons. i'm starting to test `cloud-ui` on https://www.npmjs.com/package/@hashicorp/ember-flight-icons. i'm somewhat maxed out rn, but if you want to meet the week of sept 13 and pair, i could try to make a POC on `atlas` with the package before then?

amyrlam:
threw 9/13 on your calendar, feel free to move / decline

amyrlam:
@johncowen i broke out svgo into a separate issue: https://github.com/hashicorp/flight/issues/144. and let's def talk more before next friday's meeting, i will have time to do some stuff in code before then, even a quick POC 

right now the docs are kind of behind what we've been doing in external repo code consuming these addons, so this is just to catch up and get on the same page. it's def a living document

amyrlam:
seems like some of this design documentation re: choosing sizes is live now here https://flight-staging.vercel.app/design and we could close the PR and add more as needed? or do you have more in mind @didoo @heatherlarsen?

amyrlam:
I feel like this can be closed as stale and tracked with https://github.com/hashicorp/flight/issues/133 or https://app.asana.com/0/1200639100898494/1200931531471284 instead, please reopen if I'm wrong.

PR #126, opened 2021-08-30
fix: changes height and width to relational units

MelSumner:
## :pushpin: Summary

If merged, this PR changes the `height` and `width` attributes to be `1em` so they are relational to their context. Fixes #30 

See https://codepen.io/melsumner/pen/1143304b4933a757d0a1e8233881bba7?editors=1100 for comparisons.

Examples: 

- changes component icon template
- updates tests
- updates dummy application template
- updates dummy application styles

### DIFF
--- a/ember-flight-icons/addon/components/flight-icon.hbs
+++ b/ember-flight-icons/addon/components/flight-icon.hbs
@@ -4,10 +4,10 @@
   aria-hidden="true"
   data-test-icon
   fill="{{this.color}}"
-  height="{{this.size}}"
+  height="1em"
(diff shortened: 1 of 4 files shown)


MelSumner:
Note: we shouldn't merge this until @dizzyup has had time to review. I think this resolves the issue, I just needed some time to wrap my head around the purpose of the icons and how viewBox was separate from height/width.

MelSumner:
After discussion with the group, we are on the same page that only the 24 or 16 icons should be used, so abs units only, not relational. Closing this PR.

Issue #110, opened 2021-08-25
Update tests for one source of truth


amyrlam:
Have the test comment and name be combined. Then when a test fails, it's more clear what it is, the test can be renamed as needed. Making a ticket off of https://github.com/hashicorp/flight/pull/103#discussion_r695137663 for tracking cc @MelSumner

PR #127, opened 2021-08-30
fix: removes comments from test file

MelSumner:
## :pushpin: Summary

If merged, this PR removes the comments from the test file. 

Examples: 

### DIFF
--- a/ember-flight-icons/tests/integration/components/flight-icon-test.js
+++ b/ember-flight-icons/tests/integration/components/flight-icon-test.js
@@ -6,26 +6,22 @@ import { hbs } from 'ember-cli-htmlbars';
 module('Integration | Component | flight-icon', function (hooks) {
   setupRenderingTest(hooks);
 
-  // the component should render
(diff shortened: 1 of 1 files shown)


ember-flight-icons/tests/integration/components/flight-icon-test.js
```diff
@@ -60,14 +52,12 @@ module('Integration | Component | flight-icon', function (hooks) {
       fill: 'rgb(186, 34, 38)',
     });
   });
-  // the component should have `color` set to `currentColor` (black) by default
   test('the fill should be set to black by default', async function (assert) {
```
amyrlam, line 55:
```suggestion
  test('the fill color should be set to black by default', async function (assert) {
```

suggestion to be consistent with test name below

amyrlam (approved):
thanks for fixing this!

amyrlam:
Resolves: https://github.com/hashicorp/flight/issues/110

Issue #121, opened 2021-08-30
Create routes for engineering docs vs design docs


Dhaulagiri:
In order to support the ability to surface docs for different consumers (Design/eng), create multiple routes users can click between. We'll then push documentation into those routes in subsequent issues.

PR #128, opened 2021-08-30
feature: added routes for design and engineering documentation

MelSumner:
## :pushpin: Summary

If merged, this PR resolves issue #121 

Examples: 

- added `design` and `engineering` routes to dummy app
- added navbar component to dummy app
- added styles to support navbar to dummy app
- added an `index` template to dummy app
- moved some of the application template code to the index template 
- moved controller code from application controller to index controller
- moved the router code from the application router to the index router

Needs more styling for the navbar, not sure what would be consistent and cycles are waning for today. Will pick it up again tomorrow.


amyrlam (approved):
lgtm!

would be cool to have the docs on the same page as the search down the road for a better DX

PR #132, opened 2021-08-31
Add link to the micro-website in the README file

didoo:
## :pushpin: Summary

Added a link to the Flight mini-website in the README file (always handy to have it there).

Examples: 


amyrlam (approved):
:shipit:

PR #135, opened 2021-08-31
Remove local linking from ember-flight-icons

amyrlam:
## :pushpin: Summary

We are publishing beta npm addons now, which is less confusing

cherry-picked from https://github.com/hashicorp/flight/pull/119

Examples: 


Dhaulagiri (approved):
🤷‍♂️  I think knowing this is possible still has value as someone _may_ want to play with the addon locally in their app, but I also think knowing how to do this with a published package might be/is common knowledge?

MelSumner (approved):
Totally fine with this change, but we should get folks up to speed on using `yarn link`, so it's not confusing. As an addon developer, I use this a lot- I can't imagine having to wait to have an npm package published first. That would really slow me down a lot.

amyrlam:
makes sense, how about we leave it in then and i close this PR? i'm about to update the contributing for both anyways

amyrlam:
it personally took a long time for me to get ^^ working correctly. maybe it just seemed easier in npm elsewhere, not exactly sure. i was making the change more for pushing folks to use the addon / decouple it from https://github.com/hashicorp/flight/pull/119

so closing it sgtm, if no objections

MelSumner:
I don't think we should close it, I think we should merge it; I'm saying as a meta issue, we should all get comfortable using `yarn link` or `npm link` because as addon devs we'll likely use it quite often.

amyrlam:
gotcha. ok, well i'm not fluent in it, so seems like keeping the instructions up there doesn't hurt? i'll close it for now and make another PR updating both CONTRIBUTING's which relates to heather's RFC. hope this is an ok step forward

Issue #125, opened 2021-08-30
Publish new packages and test new flow


amyrlam:
Follow-up from https://github.com/hashicorp/flight/pull/118#issuecomment-908574538

PR #118, opened 2021-08-28
Remove npm tarball and rework npm organization

amyrlam:
## :pushpin: Summary

We want `@hashicorp/ember-flight-icons` to consume `@hashicorp/flight-icons`. But for now, this is a step in the right direction that makes the npm process slightly less confusing. Happy to walk through or pair on the process with anyone also!

Review commit-by-commit, but the second commit is so large, once you get to the .svg's you kinda have the gist of it.

### :speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

Examples: 
- issue (ux,non-blocking): These buttons should be red, but let's handle this in a follow-up.
- suggestion (non-blocking): Let's change this wording to make it easier to understand.
- issue (blocking): We shouldn't introduce this kind of tech debt; let's pair and resolve the issue in a more sustainable way.

## :link: External Links

Resolves: https://github.com/hashicorp/flight/issues/108

## 👉 Next steps

- [ ] TODO: Publish a new version and test to make sure things are working well
- [ ] TODO: File any follow-ups to this task 
  - I think ideally we want @hashicorp/ember-flight-icons to consume @hashicorp/flight-icons
  - Stuff could probably be more automated with GitHub actions

Issue #90, opened 2021-08-23
Namespace icons


randallmorey:
Currently, icons live under `export-go/dist/`.  While it's possible to namespace icons in this location, it would be ideal if they were already packaged in a folder named `flight` (e.g. `../dist/flight/`).

Boundary would like to consume and use namespaced icons.

amyrlam:
following up from slack, right now the npm tarball https://github.com/hashicorp/flight/pull/99/commits/d0c4a42a5f39efec7f1074cc1e8cb1a13de6b544 has all of the .svg's in a top level folder

we should have the npm of @hashicorp/flight-icons be committed to the repo for clarity

amyrlam:
AL: Make a follow-on PR to https://github.com/hashicorp/boundary-ui/pull/720/files that uses 0.0.4-beta when published

amyrlam:
closing as done. didn't realize that "done" does not close the issue

Issue #84, opened 2021-08-20
How to handle the dupe svgs from export-go to flight-icons


amyrlam:
The svg's from the Go export are duplicated in the Ember icon. Right now they are out of sync from the latest Figma export, I manually copied them in https://github.com/hashicorp/flight/pull/69.

Some ideas:
- Have Ember addon consume https://www.npmjs.com/package/@hashicorp/flight-icons. Right now I've been publishing the flight-icons addon on a not for merge PR (https://github.com/hashicorp/flight/pull/69). So we should get that sorted on main if good to go.
- ~Document manually copying the svgs from the Go export to ember-flight-icons/public/icons directory~ update: done

Dhaulagiri:
As described option 1 seems most reasonable. The missing bit we need to solidify though is where does that actual sprite generation take place. An alternative to consider would be to have a build step available in the export node library that takes all the svgs and builds a sprite and then copies that file into the ember addon's public folder for use there.

Dhaulagiri:
See also https://github.com/hashicorp/flight/issues/48

MelSumner:
I was reading through the source code of `ember-mdi`, and it seems like that addon imports the mdi icons for use; this also allows for the configuration of only including the ones that you need. 

This could be an approach we consider.

amyrlam:
from standup: change the export-go folder to put the icons in flight-icons/, if we make that change in https://github.com/hashicorp/flight/pull/118

amyrlam:
(not urgent) right now i'm having trouble running the go script, so i can't test moving the folder:

```
flight (main) 🌘 ~ pwd
/Users/amyrlam/src/hashicorp/flight/export-go
flight (main) 🌘 ~ go run .
2021/09/01 15:49:40 Taking flight (development)
2021/09/01 15:49:40 Querying Figma file
2021/09/01 15:49:40 Error getting icons from Figma document
exit status 1
```

depends on https://github.com/hashicorp/flight/pull/138 so can revisit next week or beyond!

didoo:
After discussion with @amyrlam decided to close it, will be tackled in #28 and followup epic.

PR #138, opened 2021-09-01
Make updates to docs and commit new versions of packages

amyrlam:

Examples: 

## :link: External Links

Resolves: https://github.com/hashicorp/flight/issues/26
Resolves: https://github.com/hashicorp/flight/issues/125
Tested @hashicorp/ember-flight-icons works here: https://github.com/hashicorp/design-system-playground/pull/4
Tested @hashicorp/flight-icons works here, but with /icons in namespacing: https://github.com/hashicorp/hashicorp-www-next/pull/2267 ... this approach allows us to separate the /icons directory from the package.json, the README, and a future CONTRIBUTING


amyrlam:
@heatherlarsen this should resolve the action item you needed from me in DS-011 ... but ... since some of "how to contribute" is in question (like https://github.com/hashicorp/flight/pull/119), I'm wondering if we should just direct folks to the README

aside: i noticed that we currently link to figma in open source https://github.com/hashicorp/structure-icons#updating-icons and wondering if we can do that as well?

CONTRIBUTING.md
```diff
@@ -0,0 +1,108 @@
...
+
+After the change is merged to `main`, from the `flight-icons/` directory, run:
+
+```bash
+npm publish
+```
+
+You will need 2FA on your npm account to publish.
+
+## How to test local addon changes
```
amyrlam, line 73: from yesterday's pr discussion, i moved this for contributing, since it seems like something we should be fluent in (wasn't straightforward for me last time i tested it)

CONTRIBUTING.md
```diff
@@ -0,0 +1,108 @@
...
+npx spritesh
+```
+
+The `spitesh` command will great a `sprite.svg`. Copy this file to ember-flight-icons/
+
+```bash
+cd .. && cp flight-icons/icons/sprite.svg ember-flight-icons/public/icons && cp flight-icons/icons/_catalog.json ember-flight-icons/public/icons
+```
+
+## npm instructions
```
amyrlam, line 39: with this PR, i've already published the packages to test that the instructions were correct (before merge to `main`, took that liberty since we are on beta and needed a lot of debugging to get the instructions right). it would be cool if we had someone else follow them next publish, so that we can see what does and does not make sense!

ember-flight-icons/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hashicorp/ember-flight-icons",
-  "version": "0.0.1-beta",
+  "version": "0.0.2-beta",
```
MelSumner, line 3: @amyrlam question: I don't usually bump version numbers (or at least, I don't do it until I am going to do a release). Should we be bumping patch versions with every PR?

heatherlarsen:
@amyrlam 

> i noticed that we currently link to figma in open source https://github.com/hashicorp/structure-icons#updating-icons and wondering if we can do that as well?

From what I've heard we shouldn't be doing this, if the repo is public, but it's ok if the repo is internal or private.

ember-flight-icons/package.json
amyrlam: this bump was to test the changes from https://github.com/hashicorp/flight/pull/118#issuecomment-908574538, where we reworked the file structure

in this PR, we should commit the bumps since that's what's live in npm

definitely don't need to bump in the future unless there's an icon change, but to test that the instructions worked (https://github.com/hashicorp/flight/pull/138#discussion_r700513584) it made the most sense

amyrlam:
@heatherlarsen gotcha... yeah... i'm wondering if we can find out more about this from IT / security. i made this issue https://github.com/hashicorp/design-systems-team/issues/29

flight-icons/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hashicorp/flight-icons",
-  "version": "0.0.6-beta",
+  "version": "0.0.8-beta",
```
amyrlam, line 3: cc @randallmorey there's no change to the svg's in this version bump, but it moves the icons into a folder `@hashicorp/flight-icons/icons/`. i made that change for now, because we may want the `export-go/` to dump into that folder (this ticket: https://github.com/hashicorp/flight/issues/84), while keeping other files such as README/CONTRIBUTING in root `flight-icons/`

just FYI, there's no reason to bump your package rn, but the namespacing could change in the future. i know you mentioned that before in https://github.com/hashicorp/flight/issues/90

it could also be called `dist/` instead of `icons/`, is that generally more common? just curious

CONTRIBUTING.md
```diff
@@ -0,0 +1,108 @@
...
+```
+
+(If necessary, run a `yarn unlink`.)
+
+- In your external repo, e.g. https://github.com/hashicorp/design-system-playground-fastboot, run `yarn link "@hashicorp/ember-flight-icons"`
+- In your external repo, manually add the path to the `package.json`. For example:
+
+```json
+"devDependencies": {
+  "@hashicorp/ember-flight-icons": "link:~/your-path-here/flight/ember-flight-icons",
```
Dhaulagiri, line 91: now that the package is actually published this nuance of linking to a local path shouldn't be necessary and having it referenced like "normal" should work:

```
"@hashicorp/ember-flight-icons": "0.0.1",
```

Dhaulagiri (approved):
Made one small nit comment but otherwise this seems good to me. I suspect we might have to continue to work on _where_ some of this verbiage lives but seems 👍  to get this version of it committed

flight-icons/package.json
randallmorey: What if it were named `flight-icons/` for simpler namespacing?  I know it's redundant.  It just makes it that much easier to import.  With a folder named just `icons/`, we'd have to use either `flight-icons/icons/*` or `icons/*`, but the latter is too generic to be useful.

amyrlam: thanks - sounds good, that makes sense. or, i can explore using `files` again? i'll extract this issue to https://github.com/hashicorp/flight/issues/142 and test separately

CONTRIBUTING.md
amyrlam: was kinda leaving this in there for reference...this was in there more for testing local changes to the addon, before publish.

amyrlam: i'll leave it in for now, but we can def change this

Issue #102, opened 2021-08-24
Explain GitHub Actions test suite only fires at ember-flight-icons/ dir


amyrlam:
Test seem to be stalled on https://github.com/hashicorp/flight/pull/100

Maybe related to implementing vercel?

![image](https://user-images.githubusercontent.com/1372946/130657853-305251ec-e7d6-4b6a-a26b-1820c5c1014f.png)

Dhaulagiri:
This is because they are [configured](https://github.com/hashicorp/flight/blob/5299cc8b81e7fd88b53f2a3390bffc60c1bdd7cc/.github/workflows/ci-ember-flight-icons.yml#L44) to only run if something in the ember addon changes. This is pretty confusing and breaks the ability to require them to pass though...

amyrlam:
oh i see, thanks for pointing that out. never seen that before in tests, but haven't worked on too many addons.

should we consider having export-go output to `ember-flight-icons/`, or is that not the addon way?

Dhaulagiri:
I don't know that we'd want to consider something like that just to get the tests to behave a certain way. It'd be more a question of how we want the icon workflow to behave between figma/github.

amyrlam:
yeah makes sense, i meant more to not have 2 copies of icons. but is a tangent, will keep the issue scoped!

for this, i think i can at least add something to the readme or PR template that explains this? will propose something in a PR

PR #139, opened 2021-09-01
[🚧 WIP] Test GitHub Actions running on every PR

amyrlam:
## :pushpin: Summary

Before, tests didn't run if you didn't edit files in `ember-flight-icons/`. This was confusing to me, and would probably be confusing to external contributors.

Examples: 

## :link: External Links

Resolves: https://github.com/hashicorp/flight/issues/102

### DIFF
--- a/.github/workflows/ci-ember-flight-icons.yml
+++ b/.github/workflows/ci-ember-flight-icons.yml
@@ -30,18 +30,11 @@
 
 name: CI - Ember Flight Icons
 
-defaults:
-  run:
-    working-directory: ember-flight-icons
 on:
   push:
     branches:
       - main
-    paths:
-      - 'ember-flight-icons/**'
-  pull_request:
-    paths:
-      - 'ember-flight-icons/**'
+
 env:
   NODE_VERSION: '10.x'
 


amyrlam:
i thought this would be really quick, checked out https://github.com/jelhan/create-github-actions-setup-for-ember-addon, i see why it's not that simple now. a bit different from travis!  

maybe an intermediate step could be to add linting to `flight-icons/`, which seems like it would be a separate github actions workflow. although not sure if that's overkill. will revisit this later

amyrlam:
for context: i was looking for low-hanging fruit to pick off before i'm out

tests didn't fire at all

i think there needs to be a separate "action" for the top level repo, since this one was created (maybe?) from the travis to github actions file automated thing? i started to look at their docs and #tech-github-actions on slack

.github/workflows/ci-ember-flight-icons.yml
```diff
@@ -30,18 +30,11 @@
 
 name: CI - Ember Flight Icons
 
-defaults:
-  run:
-    working-directory: ember-flight-icons
```
Dhaulagiri, line 35: Took another pass at this in https://github.com/hashicorp/flight/pull/153. I think removing this line might have caused problems in this PR as this is still required for the tests to work correctly

amyrlam: ah gotcha - thank you for the follow-up!

PR #140, opened 2021-09-02
move conventional comments prompt to bottom

Dhaulagiri:
## :pushpin: Summary

While filling out a PR with this template it's distracting having a large section of text that is for the reviewer mixed in with content that the PR author needs to fill out. To try to help with this I propose we move that section to the. bottom as the last thing a reviewer would read before looking at the PR.

Examples: 


amyrlam (approved):
lgtm, @MelSumner cool to merge?

PR #143, opened 2021-09-02
fix: update displayed name

MelSumner:
## :pushpin: Summary

If merged, this PR changes the display name on the icon to only show the name

Examples: 

### DIFF
--- a/ember-flight-icons/tests/dummy/app/templates/index.hbs
+++ b/ember-flight-icons/tests/dummy/app/templates/index.hbs
@@ -26,7 +26,7 @@
               class="demo-icon"
             />
           </div>
-          <p>{{meta.name}}</p>
+          <p>{{Onlyname meta.name}}</p>
         </li>
       {{/each}}
     </ul>


didoo (commented):
issue (blocking): we're removing a vital piece of information

As a developer, is useful to have the full-name of the icon so that I can copy it from the website UI and paste it directly in the code. I think what we show under the icon should be exactly the name of the icon to use.

Dhaulagiri:
My comment spurred this so I don't want to cause us to spin our wheels _too_ much on this. I think we're all correct depending on the context:

* If I'm trying to use the ember addon, seeing `activity-16` in the docs requires me to know to drop `-16` when I use `<FlightIcon @name="activity" />`. We saw this happen during our waypoint pairing session today.
* If I'm reading the docs as currently organized, it's potentially confusing to have two icons next to each other with the same name.

didoo:
OK, now I see the issue. In Ember we're coupling the name of the glyph (eg. `activity-16`) with the actual geometrical size of the element in the UI:

```
<svg 
  class="flight-icon icon-{{@name}} {{if this.display "display-inline"}}"
  ...attributes
  aria-hidden="true"
  data-test-icon
  fill="{{this.color}}"
  height="{{this.size}}"
  id={{this.iconId}}
  viewBox="0 0 {{this.size}} {{this.size}}"
  width="{{this.size}}" 
  xmlns="http://www.w3.org/2000/svg"
>
  <use href='{{this.contextRootURL}}@hashicorp/ember-flight-icons/icons/sprite.svg#{{@name}}-{{this.size}}'></use>
</svg>
```

This doesn't work (sorry for noticing only now, I am still getting my head around the APIs of the FlightIcon component).
This is under the assumption that we always want a 16px icon in a 16px box and a 24px icon in a 24px box, but this does not cover all the possible cases. In all the other cases where the icon is not one of the standard sizes, it would not work. 

What if we need a 12px icon size ( but the glyph/symbol name in that case should be `name-16`)? What if we need a 20px icon size ( but the glyph/symbol name in that case should be `name-24`)? What if we need a 100% icon size ([we discussed it yesterday](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1630595366003000))? 

Strong opinion here, based on experience: the icon name and the icon geometrical size should be definitely different (and independent) props/entities. 

Let's discuss maybe in a dedicated meeting? (I have also some doubts about other defaults of the `.hbs` template)

There are other things in that API/defaults that I would like to discuss,

didoo:
ok, after discussion with @dizzyup I wasn't able to convince him of why we should support the component (FlightIcon) to support also a "stretch" sizing, and he preferred to have smart defaults. So let's forget about this. 
If I was wrong, better so. If I was right, we'll learn something.

Dhaulagiri:
I prompted this without much consideration to the other possibilities on how we might better organize this information. I think we could close this PR and revisit as part of the overall design polishing effort we'll do as we push toward 1.0.

MelSumner:
Agree. Closing.

PR #149, opened 2021-09-03
enhancement: improved engineer docs

MelSumner:
## :pushpin: Summary

If merged, this PR adds additional information to the docs for engineers. Includes links at the top of the page types of use.
See: https://flight-git-feature-update-eng-docs-hashicorp.vercel.app/engineering specifically.

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

Examples: 
- issue (ux,non-blocking): These buttons should be red, but let's handle this in a follow-up.
- suggestion (non-blocking): Let's change this wording to make it easier to understand.
- issue (blocking): We shouldn't introduce this kind of tech debt; let's pair and resolve the issue in a more sustainable way.

PR #152, opened 2021-09-03
bugfix: resolve issue where not all the css classes were namespaced

### LINKED ISSUE #150: bug: update css classes to avoid potential naming collisions
Not all of the css classes in the flight icon are prefixed properly which may cause situations that result in naming collisions. 

To fix, 
- [x] update the classes on the SVGs with flight- prefix.
- [x] update the component styles
- [x] update the documentation
- [x] make sure the tests still pass

MelSumner:
## :pushpin: Summary

If merged, this PR resolves #150 by adding additional namespacing to the component css classes. I also re-arranged the properties in the component itself so height/width/viewbox were grouped together, for easier maintenance (based on feedback from @didoo). 

After this PR and PR #149 are merged, I'll update the documentation and the README files to align.

Examples: 

### DIFF
--- a/ember-flight-icons/addon/components/flight-icon.hbs
+++ b/ember-flight-icons/addon/components/flight-icon.hbs
@@ -1,13 +1,13 @@
 <svg 
-  class="flight-icon icon-{{@name}} {{if this.display "display-inline"}}"
+  class="flight-icon flight-icon-{{@name}} {{if this.display "flight-icon-display-inline"}}"
   ...attributes
   aria-hidden="true"
   data-test-icon
   fill="{{this.color}}"
-  height="{{this.size}}"
   id={{this.iconId}}
-  viewBox="0 0 {{this.size}} {{this.size}}"
+  height="{{this.size}}"
   width="{{this.size}}" 
+  viewBox="0 0 {{this.size}} {{this.size}}"
   xmlns="http://www.w3.org/2000/svg"
 >
   <use href='{{this.contextRootURL}}@hashicorp/ember-flight-icons/icons/sprite.svg#{{@name}}-{{this.size}}'></use>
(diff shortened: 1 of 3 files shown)


ember-flight-icons/addon/components/flight-icon.hbs
```diff
@@ -1,13 +1,13 @@
...
-  class="flight-icon icon-{{@name}} {{if this.display "display-inline"}}"
+  class="flight-icon flight-icon-{{@name}} {{if this.display "flight-icon-display-inline"}}"
   ...attributes
   aria-hidden="true"
   data-test-icon
   fill="{{this.color}}"
-  height="{{this.size}}"
   id={{this.iconId}}
-  viewBox="0 0 {{this.size}} {{this.size}}"
+  height="{{this.size}}"
```
didoo, line 8: nitpick: should the width come before the height?

I've always seen the width before the height (eg. https://www.w3.org/TR/css-sizing-3/#preferred-size-properties) and the same order happens in the viewBox (https://svgwg.org/svg2-draft/coords.html#ViewBoxAttribute)

didoo (approved):
You're a hero ;)

ember-flight-icons/addon/components/flight-icon.hbs
```diff
@@ -1,13 +1,13 @@
...
   width="{{this.size}}"
```
MelSumner, line 9:
```suggestion
  width="{{this.size}}" 
  height="{{this.size}}"
```

ember-flight-icons/addon/components/flight-icon.hbs
MelSumner: swapped! 👍

PR #153, opened 2021-09-03
internal: run ci on all changes

Dhaulagiri:
## :pushpin: Summary

Pros: makes the tests run no matter where changes triggered in the codebase and we can then require them to pass
Cons: Runs tests for code even if none of that code has changed


MelSumner (approved):
I agree with this change.

Issue #28, opened 2021-08-09
Create skeleton for node icon export library


Dhaulagiri:
### Todo
- [ ] name (`icon-export`?)
- [ ] uses yarn
- [ ] has appropriate license
- [ ] author: `HashiCorp`

didoo:
Closing this one because most of the open questions have been figured out, and with #171 we have more or less finalized the folder structure.

PR #154, opened 2021-09-06
[DO NOT MERGE] Create skeleton for node icon export library

didoo:
## :pushpin: Summary

This is an initial proposal (see also #28) for what I see should be the folders structure and files organisation (and also some high-level logic) of the Node version of the Flight icons/assets pipeline. There are still a lot of open questions, for which I don't have an answer (yet).

See the description below for more details.

**🚨 IMPORTANT 🚨 - DO NOT MERGE!**
This is intended only as a proposal to start a discussion, not an actual PR. **Don't merge!**

As mentioned above, this is a PR about the folders structure and files organisation for the Node version of the Flight icons/assets pipeline. I have moved around (and in some cases also deleted) a lot of files, but it was necessary to have a clean folders and files structure to see altogether. So don't worry, I'm not out of my mind 😀 I'll soon open a PR where you can start to review proper code changes. 

### How to review this PR.

No specific format: every comment, feedback, input, suggestion, critique is welcome.

Don't focus too much on the changes (so potentially even ignore the GitHub diff) and look just at the resulting structure.
My suggestion would be to checkout this branch in a completely new folder, so you will not destroy your current _flight repo_ setup: 

```
cd ~/Desktop && mkdir flight-temp && cd flight-temp
git clone --branch 28-node-icon-export-library-skeleton https://github.com/hashicorp/flight.git
open .
```

###  Things to look at:

My high-level vision approach has been to have three clear steps in the pipeline process: an input (`src`), a set of Node scripts (`scripts`) that process this input, and generates a set of "bundles" in output (`dist`). These bundles may be then released as NPM packages to be consumed by the web clients.

The **SRC** folder contains:

- the `flight-icons-svg` folder where the SVG files are stored as they are retrieved from Figma, without any processing, just as they are (plus a "catalog.json" file that contains the list of all the icons and their relevant metadata, at the time when they were synced)
- the `ember-flight-icons` at the moment is just the previous folder moved under the SRC parent (at the moment I don't have enough knowledge of how an Ember addon is created; the idea for the moment is to  just reuse the existing process and update it to have the DIST folder in the root as destination; in a second iteration I'll try to understand if there is space for simplification/improvement there too)

The **SCRIPTS** folder contains:

- the `.env` environment file (TBD if should be in the root)
- the `config.ts` file that contains the specs of what Figma files/pages/frames to process
- the `build.ts` script (open the file to see an outline of the logic of the file) that will be used to process the source SVG files and create the bundles in output
- the `release.ts` script (open the file to see an outline of the logic of the file) that will be used to take the generated bundles and publish them on NPM as packages
- the `package.json` file used to call the two commands `build` and `release`

The **DIST** folder at contains the output of the "build" process for the different formats that we want to "target" (at the moment we're releasing Ember Addon and SVG standalone, but we may want to release an SVG sprite as well, plus a CSS version, a React version, etc). Keep in mind this is just a suggestion to give an idea of how the `dist` folder will look like. These "bundle" folders will then be further processed and converted to packages ready to be published under NPM.

The **DOCS** folder contains some of the READMEs that we want to make available, and that possibly will be later embedded in the NPM packages for release (to be discussed/decided if this is the right place, and if we really want to release READMEs in markdown inside the NPM packages).

### Things still to decide/discuss:

1.  Is it ok for everyone to use TypeScript? (don’t know the team’s proficiency, mine is OKish but I prefer to stay on the side of using TS vs not)

2.  Name of the folder "scripts" (is it clear what it is? for me it is, but not sure for others)

3.  Do we want to release independent versions for the different outputs, or whenever we have a change we bump all the generated packages? Imagine for example we update the APIs of the Ember <FlightIcon> component but the icons themselves have not changed. _(This is in relation to the location of the package.json used for the generation of the different NPM packages)_

4. The current proposed folder structure for the DIST is made under the assumption that we will release independent packages for each format:

    - ember-flight-icons
    - flight-icons-svg
    - flight-icons-svg-sprite
    - flight-icons-css
    - flight-icons-react (this is an example)

    But an alternative could be to release just two packages:

    - ember-flight-icons
    - flight-icons
        - SVG
        - Sprite
        - React
        - CSS

    Personally I would go with the first one, also because (in reality) it's likely we will not release all these different packages!

5. micro-website location: at the moment is still under the `ember-flight-icon` folder (as far as I understand, maybe I missed something) but probably should be moved to another place (eg. root of the project, or under `/docs` folder?). plus, to understand if it "consumes" directly the assets (and the `catalog.json` or uses an NPM package as dependency).

6. probably much more, but for the moment let's start with this and then we'll go in more details


Dhaulagiri:
👏  the general shape of this looks pretty reasonable to me. Some comments below inline:

> * Is it ok for everyone to use TypeScript? (don’t know the team’s proficiency, mine is OKish but I prefer to stay on the side of using TS vs not)

I don't know that we have a lot of experience with it but this seems like a valuable tool for us to get more familiar with as we build out different projects

> * Name of the folder "scripts" (is it clear what it is? for me it is, but not sure for others)

Makes sense to me

> * Do we want to release independent versions for the different outputs, or whenever we have a change we bump all the generated packages? Imagine for example we update the APIs of the Ember  component but the icons themselves have not changed. _(This is in relation to the location of the package.json used for the generation of the different NPM packages)_

That's a good question. I think you could make a case for either option. 

> * The current proposed folder structure for the DIST is made under the assumption that we will release independent packages for each format:
>   
>   * ember-flight-icons
>   * flight-icons-svg
>   * flight-icons-svg-sprite
>   * flight-icons-css
>   * flight-icons-react (this is an example)
>   
>   But an alternative could be to release just two packages:
>   
>   * ember-flight-icons
>   * flight-icons
>     
>     * SVG
>     * Sprite
>     * React
>     * CSS
>   
>   Personally I would go with the first one, also because (in reality) it's likely we will not release all these different packages!

Will actually need a `dist` folder if the scripts are outputting packages to npm which will contain a "dist" folder with their assets? I think it would be more likely that the `scripts` folder matches this format instead?

In either case, I think the first proposal makes the most sense.

> * micro-website location: at the moment is still under the `ember-flight-icon` folder (as far as I understand, maybe I missed something) but probably should be moved to another place (eg. root of the project, or under `/docs` folder?). plus, to understand if it "consumes" directly the assets (and the `catalog.json` or uses an NPM package as dependency).

I agree we should move it as it should be pretty easy to do and will help clarify that it isn't solely for the ember addon, but doing so should be a relatively low priority.

amyrlam:
👍 TypeScript sgtm

didoo:
After discussion with @Dhaulagiri and @MelSumner today, about how to "build" and "release" the Ember addon in the context of the Node script, we have agreed that probably is better to keep the current release mode for the Ember addon (essentially is as a "consumer" of the flight-icons).

For this reason I probably will adopt a different folder structure, at this point (which is also more similar to the initial one):

- **flight-icons**: this folder will contain the elements used to generate the `flight-icons` packages (the "source" folder with the SVG exported from Figma, the `sync/build/release` "scripts" and the "dist" folder with the bundles to be released as NPM packages)
- **ember-flight-icons**: this will contain everything that is related to the Ember addon (and the mini-website); potentially as suggested by @Dhaulagiri we may even be able to "consume" the "flight-icons" package as _npm_ dependency for the Ember addon

didoo:
Closing this one because most of the open questions have been figured out, and with #171 we have more or less finalized the folder structure.

PR #147, opened 2021-09-02
adding design documentation to README

heatherlarsen:
## :pushpin: Summary

If merged, this PR adds consumer and contribution instructions for designers to the README

### :speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

Examples: 
- issue (ux,non-blocking): These buttons should be red, but let's handle this in a follow-up.
- suggestion (non-blocking): Let's change this wording to make it easier to understand.
- issue (blocking): We shouldn't introduce this kind of tech debt; let's pair and resolve the issue in a more sustainable way.

## 👉 Next steps
* Add design guidance (ie. icon types, when to use certain icons, icon geometry, etc)

PR #157, opened 2021-09-07
enhancement: updating the design documentation

MelSumner:
## :pushpin: Summary

If merged, this PR absorbs the content from @heatherlarsen PR (#147 ) and puts it in the design.hbs file. 
It also streamlines the header and puts the page title and navbar in the same row. 

Specifically, see: https://flight-git-feature-docs-design-hashicorp.vercel.app/design 

**:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.**

Examples: 


didoo (commented):
Had a quick look at the resulting website (didn't read the documentation itself) and for me is OK. Nice improvements!

Dhaulagiri (approved):

PR #159, opened 2021-09-08
Initial step to migrate docs to tailwind

Dhaulagiri:
## :pushpin: Summary

This PR lays the foundation for using tailwind for our docs. It currently does not actually import the tailwind styles at all since they break our existing design, but you can enable them for local testing and/or in subsequent PRs when we actually implement the new design.

Pull down the PR and add a tailwind class to a template and confirm it shows up correctly

## :+1: Definition of Done

- [x] Tailwind added
- [x] Determine how to address node incompatibility issue #160 
```
error tailwindcss@2.2.11: The engine "node" is incompatible with this module. Expected version ">=12.13.0". Got "10.24.1"
```

### DIFF
--- a/ember-flight-icons/ember-cli-build.js
+++ b/ember-flight-icons/ember-cli-build.js
@@ -7,6 +7,8 @@ module.exports = function (defaults) {
     // Add options here
     postcssOptions: {
       compile: {
+        enabled: true,
+        cacheInclude: [/.*\.hbs$/, /.*\.css$/, /.*\.html/],
         plugins: [
           {
             module: require('postcss-import'),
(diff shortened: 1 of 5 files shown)


MelSumner (approved):
Seems good. Is purgeCSS the second step?

Dhaulagiri:
> Is purgeCSS the second step?

Assuming I've wired things up correctly, purgeCSS is no longer necessary with the advent of [JIT mode](https://tailwindcss.com/docs/just-in-time-mode#enabling-jit-mode). Only CSS we are actually referencing should be included in production builds. We'll need to confirm this works correctly when we actually enable tailwind.

PR #160, opened 2021-09-08
drop support for node 10

Dhaulagiri:
## :pushpin: Summary

ember-cli has dropped support for this, it's been EOL for some time and I believe we can drop support for it as well as I'm not aware of any consumer apps on a node version this old (and if they are they should probably upgrade!)


MelSumner (approved):
I think I've had to put `"node": "10.* || 12.* || >= 14"` in a few of the apps I was testing the Ember addon in. As long as we're sure we won't cause compat issues, I'm all for moving forward!

Issue #161, opened 2021-09-08
Export script (Node version) - Setup


didoo:
Extracted this task from #155 so can be reviewed independently (and potentially merge in master if approved, since it doesn't have any overlapping with the existing code).

The task is about the initial setup of the Node export script in terms of TypeScript, eslint, maybe even prettier (TBD).

Issue #155, opened 2021-09-07
Export script (Node version) - Syncing


didoo:
#### Description
This is the first part (the "syncing") of the new Node.js version of the "Figma to Flight" export script.
What it does is to download the SVG assets in a specific folder, overriding the default file names (Figma uses the variant name for it, and that would be `Size=16` or `Size=24`)

#### Expected logic
This is an outline of what we intend to implement

```
INIT
- get a few options from the command line arguments [TBD, it depends on how we want to use the script]
- read a configuration file to get the list of files/pages/frames we want to sync from

SYNC
- loop over the files, pages and frames to get a list of all the icons (Figma "components") we need to export
- loop over the list of assets we need to process and for each one of them
  - get all the relevant meta-data about the asset
    - component set name (will be used later to save the SVGs with a more meaningful filename)
    - size
    - description (used to provide alternative search terms)
  - download the asset in SVG format and rename it accordingly (if in the future we may need different formats (eg. PNG for email) we'll modify the script accordingly)
- save a "catalog.json" file containing the list of assets and their relevant metadata
```

#### Things to consider/keep in mind

- We intend to use an open source library (https://github.com/marcomontalbano/figma-export or as a fallback https://github.com/tsimenis/figma-export-icons). If these libraries will not meet our needs, we'll implement a custom script.
- To fix the problem with the naming of the exported files, we'll use the REST API to retrieve the icons metadata (parent name, size, description etc). At the moment of writing, is not yet clear how this will work (and if it will work at all!): it depends on the library we're using, we need to pass an ID for the export of the Figma component so we can associate it with the metadata and rename the exported file to be meaningful).
- We'll try to implement the script as agnostic as possible on the exported format (reason: for now we use SVG, but in the future we may also need other formats, eg. PNGs for emails, PDF for slide decks, etc).
- At the moment we'll use the folder structure proposed in #28. It will be easy to change later, when it's finalised.

#### Tasks
- [x] Setup the script so that can be run via CLI (package.json + initial code)
- [x] Build a working prototype to test the existing open source libraries
- [x] Check if it's possible to require a specific ID/icon (so later we can change its name using the metadata)
- [x] If it's possible, add the logic to the script to rename the downloaded assets (and avoid conflicts/overwrites at runtime)
- [ ] ~~If it's **not** possible, build a custom script to export the assets ourselves.~~
- [x] Add logic to create the `catalog.json` file
- [x] Add README file (similar to the GO one)
[...]

PR #162, opened 2021-09-08
Export script (Node version) - Setup (PR)

didoo:
## :pushpin: Summary

Initial setup of the following tools (this solves #161):

- **[TypeScript](https://www.typescriptlang.org/)**
- **[eslint](https://eslint.org/)**

An open question is if we want to add also [prettier](https://prettier.io/), but I'd prefer to discuss with the team if we want to adopt it too (for me, yes).

_(Notice: this is a cherripick of 508a5af0d09d8d743ea01d4ae095f9e3440f5c24 from #155)_

### How to review this PR.

Look at the diff in GitHub. 

### Things to look at:

- `package.json` - is the right format? do we want to add authors? other info?
- `tsconfig.json` - is everything ok here?
- `.eslintrc.js` - is everything ok here?

## :link: External Links

Issue: #161

If you want to test the setup, pull the branch and then:

```
// install dependencies
cd ~/[project-path]/script && yarn

// test typesafety
yarn typecheck

// code linting
yarn lint

// run a mock script
yarn sync
```

## :+1: Definition of Done

- [ ] package.json correctly set-up
- [ ] TypeScript added and correctly configured
- [ ] eslint added and correctly configured
- [ ] Prettier added (optional)

Examples: 


didoo:
@MelSumner @amyrlam I have asked @Dhaulagiri to review this small PR (in the context of the larger Node export script project). If haven't added you as reviewers, but if you have time (and opinions!) on this, I would like to have your feedback (among other things, if we want to introduce also _prettier_ or not).

scripts/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "flight-icons-scripts",
+  "version": "0.0.1",
+  "private": true,
+  "main": "index.js",
+  "license": "MIT",
```
Dhaulagiri, line 6:
```suggestion
  "license": "MPL-2.0",
```

Dhaulagiri (approved):
this seems reasonable to me. I don't have a lot of experience with typescript to offer any valuable feedback, so I'd defer to your experience there.

As to the question of prettier I say yes but also that we could add it in subsequent PRs, particularly if we find a default config we like.

amyrlam:
Prettier is great for me! 

FWIW, we have Prettier in ember-flight-icons out of the box, but not in CI yet it looks like from these links:
https://github.com/hashicorp/flight/blob/main/ember-flight-icons/package.json#L47
https://github.com/hashicorp/flight/blob/main/.github/workflows/ci-ember-flight-icons.yml#L81
https://github.com/hashicorp/flight/blob/main/ember-flight-icons/package.json#L92-L100

We may want one Prettier to absorb the other, I vaguely remember doing this elsewhere.

Don't care re: Prettier defaults, a pre-commit hook can help with that if ppl disagree. Not suggesting it now just throwing in my two cents

PR #163, opened 2021-09-08
add correct tailwind imports

Dhaulagiri:
## :pushpin: Summary

The previous list was incomplete and when uncommented would not have given us access to all the classes


amyrlam (approved):
:shipit: 

are we still merging PR's at will? just checking the protocol if a reviewer isn't assigned

MelSumner (approved):
👍

PR #164, opened 2021-09-09
Export script (Node version) - Syncing (PR)

didoo:
## :pushpin: Summary

Removed the Go script and replaced it with a Node script that exports the icons from Figma, saves them in a target folder, and creates a `catalog.json` file (this solves #155):

### How to review this PR.

Look at the diff in GitHub. Unfortunately, there are a lot of changes having removed the Go script, and moved the existing SVG and `catalog.json` in the `src/flight-icons-exported` folder. Please focus on the `scripts` folder for now (you can use the "filter" utility in the "Files Changed" page to reduce the noise).

### Things to look at:

- the overall content/organization of the folder `/scripts`
- the main script`/scripts/sync.ts` and its sub-modules in the folder `/scripts/sync-parts`
- the `config.ts` file
- the type definitions

### Things that are still missing/incomplete:

- the SVGs and the catalog are the old ones, I'll run a sync as soon as I have cleared with @dizzyup why I see some changes compared to production
- ~~the README about how to use and update the script will be added later~~
- the actual "build" step that generates the different bundles (for release) will be done later

At the moment, the local version needs some manual setup to work (because of this issue https://github.com/marcomontalbano/figma-export/issues/92). In case you want to test it locally I can guide you in how to do it.

## :+1: Definition of Done

See #155

Examples: 


scripts/config.ts
```diff
@@ -0,0 +1,16 @@
+import { ConfigData } from "./@types/ConfigData";
+
+export const config: ConfigData = {
+    // the data of the Figma file we sync with
+    figmaFile: {
+        // this is the "production" file
+        id: 'TLnoT5AYQfy3tZ0H68BgOr',
```
amyrlam, line 7: non-blocking question: do we want this to be a .env secret instead?

scripts/README.md
```diff
@@ -0,0 +1,104 @@
...
+The synchronization step is executed by the `/scripts/sync.ts` file. To run this script use the following command in your CLI:
+
+```bash
+yarn sync
+```
+
+This action will:
+
+* Retrieve the assets metadata from Figma via REST API
+* Export the assets as `.svg` files into `/src/flight-icons-original/svg`
```
amyrlam, line 24: question (non-blocking): 

i had tried to have the go export simply dump the files into https://github.com/hashicorp/flight/tree/main/flight-icons/icons (we could rename that directory to `dist` or `svg` or `src`, no opinion) but i couldn't get the exporter to traverse the tree, maybe it was broken though

maybe we can explore that after this and update the instructions of all the manual munging separately https://github.com/hashicorp/flight/blob/main/CONTRIBUTING.md#figma-export-go-instructions

amyrlam (approved):
i had taken a look yest - i'm not a node expert but it lgtm, happy to merge it in so can test it out! 

left a couple non-blocking questions, happy to address later

amyrlam:
p.s. i'll let you merge it in

scripts/config.ts
didoo: why? it's not a secret thing. is an ID of a Figma file (that has its own permissions, so even if a user knows the ID, they can't do anything with it). 

@Dhaulagiri any thoughts on this?

scripts/README.md
didoo: as far as I understand, the go exporter is expecting a very specific node/nesting structure, which in reality can be changed. this script doesn't care of where the icons are in the tree, because is using the "export" provided by the REST API directly.

Re. the readme, if you see in the /scripts/README you can see the new/updated instructions of how to run the "sync" script.

didoo:
@Dhaulagiri see comments above. let me know if it can be merged also for you.

scripts/config.ts
Dhaulagiri: we had a bit of back and forth on this before you joined and while we did move it to to an `.env` we ultimately agreed with what you are saying here. If it doesn't seem like something needs to be configurable I think this is ok.

scripts/README.md
amyrlam: sounds good, let's keep it as-is for now!

Dhaulagiri (approved):

PR #174, opened 2021-09-09
allow embroiderSafe to fail

Dhaulagiri:
Known issue with ember-cli-postcss https://github.com/embroider-build/embroider/issues/688

I think it would be worth consdiering either not running this step at all since allowing it to fail is effectively the same thing or extracting out the docs so issues with them don't affect our ability ensure embroider compat for consumers of the addon itself.


MelSumner (approved):
👍

Issue #176, opened 2021-09-09
update docs to read new catalog.json format


Dhaulagiri:
As part of the export work the format of catalog.json will be changing. We need to update the docs site to

- [x]  read this format instead (and if the format isn't working feed that back into the export work @didoo is doing)
- [x] Search also on the values in `description`

```
{
  "lastRunTimeISO": "2021-09-09T15:53:34.232Z",
  "lastRunFigma": {
    "id": "TLnoT5AYQfy3tZ0H68BgOr",
    "page": "Export",
    "frames": [
      "Core",
      "Services"
    ]
  },
  "assets": [
    {
      "id": "350:23",
      "fileName": "activity-24",
      "iconName": "activity",
      "description": "activity, pulse, health",
      "size": "24",
      "width": 24,
      "height": 24
    },
    {
      "id": "350:25",
      "fileName": "activity-16",
      "iconName": "activity",
      "description": "activity, pulse, health",
      "size": "16",
      "width": 16,
      "height": 16
    },
    {
      "id": "350:31",
      "fileName": "alert-circle-24",
      "iconName": "alert-circle",
      "description": "alert, circle, error, attention",
      "size": "24",
      "width": 24,
      "height": 24
    },
...
  ]
}
```

**Note: We can open a PR validating this without but wait to merge it until the export is finished**

didoo:
@Dhaulagiri it may be that we need to include the new assets (otherwise the catalog will be out of sync with the SVGs in the repo)

didoo:
now that #183 has been merged, this one is ready to be picked up (probably @MelSumner next week?)

Dhaulagiri:
See #179

PR #179, opened 2021-09-09
ingest and search on new catalog format

Dhaulagiri:
## :pushpin: Summary

Closes https://github.com/hashicorp/flight/issues/176

This PR uses the new catalog.json format and allows us to search against its added description field. I also generate a new sprite based on the new set of icons which could be pulled into a separate PR if that would be helpful.

## :camera_flash: Screenshots

Screen Shot 2021-09-09 at 15 20 29

Visit the preview app and search for icons both based on name and on the `description` values

## :+1: Definition of Done

- [x] New functionality works 
- [x] new catalog.json format actually exists and we can stay in sync with the icons that exist https://github.com/hashicorp/flight/pull/183

Examples: 


amyrlam (approved):
:shipit: lgtm

PR #180, opened 2021-09-09
add simple acceptance test to confirm docs are rendering icons

Dhaulagiri:
## :pushpin: Summary

The goal of this PR is to give a simple test to confirm that the docs are rendering at least one icon in the list of icons. This will help guard against any unintended breakage as we refactor how we are ingesting the catalog.json file.


MelSumner (approved):
Good idea!

PR #182, opened 2021-09-10
enhancement: add toggle button

### LINKED ISSUE #169: Implement new icon size toggle functionality
Screen Shot 2021-09-08 at 20 01 17

Provided this design is reasonably straightforward to implement, toggling the button should switch between rendering 16px and 24px icons.

MelSumner:
## :pushpin: Summary

If merged, this PR resolves #169 by adding the toggle button for the icon size.

## :camera_flash: Screenshots

So far: 
image
image

Examples: 


MelSumner:
Looks like I need a rebase too, I'll keep working on this tomorrow.

MelSumner:
@Dhaulagiri I need to hook up the toggle to the icon grid but as a standalone MVP it does what it says on the tin.

Dhaulagiri (approved):

PR #184, opened 2021-09-10
Removed entirely the “export-go” folder

didoo:
## :pushpin: Summary

This is just a follow up of task #155, in which we didn't remove the folder to avoid too much noise in the code diff.

On a side node, the script was not working anymore after the recent changes in the layer naming/structure/nesting in Figma, so... in any case it was useless at this point.

screenshot_57


Dhaulagiri (changes requested):
Can we also remove these other references to export-go?

* `.tours`
*  the workflows `export-go.compile.yml` and `export-go.main.yml`
* [Instructions in the contributing guide](https://github.com/hashicorp/flight/blob/main/CONTRIBUTING.md#figma-export-go-instructions)

didoo:
@Dhaulagiri nice catch, makes sense. pushed the changes

Dhaulagiri (approved):

PR #183, opened 2021-09-10
Add newly generated SVGs (and their catalog) to the repository

didoo:
## :pushpin: Summary

This is just a follow up of task #155, in which we didn't add the SVGs to avoid too much noise in the code diff.

PR #186, opened 2021-09-10
Regenerate assets after fixing issue with “Clip content” in Figma

didoo:
## :pushpin: Summary

I have re-generated the assets after fixing an issue with Figma, where the icons had "Clip content" selected and this was causing Figma to add extra code (in particular an extra "ClipPath") to the exported SVGs (see example below) when the glyph/content of the icon/component was touching the borders of the bounding box.

https://github.com/hashicorp/flight/pull/183/files#diff-a9bff655809f99647a1cb9bda4d6977a21f0461b147599659ed5f83ef047050f

The solution is to uncheck the "Clip content" flag in the Figma UI for that element (I did it via scripting).

/cc @dizzyup @heatherlarsen can you please update the documentation for the designers (in Figma and in the mini-website) to make sure this is explained. Thanks.

-----

For future reference: 

How I understood what was causing the problem:
https://forum.figma.com/t/setting-an-explicit-svg-viewbox/2504/7

Code used to fix the issue in Figma:

```JavaScript
var allComponents = figma.currentPage.findAll(node => node.type === "COMPONENT");
if (allComponents.length > 0) {
    console.log(`Found ${allComponents.length} components.`);
    allComponents.forEach((node) => {
        const currName = node.name;
        if (currName.match(/^size=(?:16|24)$/i) || currName.match(/^size=(?:16|24), style=(?:color|mono)$/i)) {
            // actually is easier that one can think, is just a matter of renaming the layer!
            node.name = currName.replace(/style/i,'tint').toLowerCase();
            console.log(node.name);
        }
    });
} else {
    console.log('No components found!');
}
```


didoo:
@Dhaulagiri before merging, can you please regenerate the SVG sprite and commit it to this branch? thanks

(next week we'll merge also the build part, and this will be done automatically)

Dhaulagiri:
Is the fix for this something we could automate as part of the sync script just in case this comes up again on the figma side?

Dhaulagiri (approved):

didoo:
> Is the fix for this something we could automate as part of the sync script just in case this comes up again on the figma side?

We could have a Figma plugin that does it automatically, I can create it in 5 mins. But I think is not useful (I can explain why).

PR #189, opened 2021-09-13
don't bother running tests we are allowing to fail

Dhaulagiri:
## :pushpin: Summary

with github ci there is no real value to running tests that we allow to fail because you would have to manually inspect each one to figure out if the test passed or not. Since we are ok with them not passing we might as well not run them at all and save 3 run's worth of resources.


amyrlam (approved):
:shipit:

PR #191, opened 2021-09-13
added copy to clipboard functionality for code samples

MelSumner:
## :pushpin: Summary

 If merged, this PR adds functionality to copy some of the code samples to the user's clipboard

- used [ember-prism](https://github.com/shipshapecode/ember-prism)

## :camera_flash: Screenshots

image

Examples: 


MelSumner:
H/T to @heatherlarsen for this idea. 👍

Dhaulagiri (approved):

PR #192, opened 2021-09-13
enhancement: add code highlighting

MelSumner:
## :pushpin: Summary

If merged, this PR should fix some prettier/formatting issues and also adds code highlighting.

## :camera_flash: Screenshots

image

Examples: 


MelSumner:
Note: I'd like to have linting done automatically on push, that way I'm prevented from pushing code if it won't pass the linter.

Issue #171, opened 2021-09-09
Export script (Node version) - Building


didoo:
#### Description
This is the second part (the "building", or "bundling"?) of the new Node.js version of the "Figma to Flight" export script.
What it does is to take the exported assets, process them, and generate multiple "bundles" in different formats (eg. standalone SVGs, a single SVG sprite, an Ember Addon, SVGs embedded as base64 in a CSS/Sass file). These bundles will be then used in the "release" step to generate the NPM packages and release them to the registry.

#### Expected logic
This is an outline of what we intend to implement

```
- create a TEMP directory (will be ignored anyway by git)
- copy the SVG folder in the TEMP directory
- optimize the SVG files via SVGO (if we need different optimizations for different targets, we'll create different folders in the TEMP directory)
- empty the DIST directory (if already exists)
- using the optimized SVGs, generate the files required by the the different "formats" that we want to distribute (the one below is just an example) and save them in the DIST folder
  - SVG - these will be plain SVG files
  - SVG sprite - this will be later consumed by the Ember Addon (or potentially as standalone file)
  - Ember Addon - same as the existing one
  - CSS/SASS - this needs to be agreed upon with John Cowen (Consul)
- delete the TEMP directory
```

#### Things to consider/keep in mind

- Nothing at the moment, I'll update this list later.

#### Tasks
- [x] Setup a process that takes the source SVG files, and:
    - [x] process them (removing fills so will inherit color via `currentColor`)
    - [x] optimize them
    - [x] saves them in a "temp" folder
- [x] Add a "build" step that copies these SVG files in the DIST folder
- [x] Setup a process that takes the optimized SVG files, creates an SVG sprite (with unique IDs), and saves it in a "temp" folder
- [x] Add a "build" step that copies this SVG sprite in the DIST folder
- [x] Pair with Brian or Melanie and decide what to do with the Ember Addon process (and what is the "src" for it)
- [x] Add a "build" step that updates the Ember Addon with the new version of the catalog.json and the SVG sprite
- [ ] ~~Add a "build" step that takes the SVGs and embeds them in a CSS/SASS and saves it in the DIST folder~~
    - see https://github.com/hashicorp/flight/issues/197
- [x] Update the README file with the documentation for the "build" steps

PR #193, opened 2021-09-13
Regenerate assets after splitting between “mono” and “color” icons in Figma

### LINKED ISSUE #185: Fix how colored icons are shown in the miniwebsite
In #179 we've added the newly generated SVGs to the mini-website. They now contain also the colored icons, but for some reasons the mini website is using the mono icons also for the colored ones. See screenshot below:

screenshot_67

didoo:
## :pushpin: Summary

This should fix #185 and avoid us to change the API of the `FigmaIcon` component to support "tinted" icons (now the icon "name" contains an extra `-color` suffix when is colored, so effectively is a distinct icon, not a variant.

Please @Dhaulagiri can you check if this fix #185? thanks


Dhaulagiri:
This seems to mostly have fixed the problem with the exception of `azure-devops-color` which I can investigate tomorrow

didoo:
the issue with the `azure-devops-color` icon is that there are conflicting IDs in the sprite (probably the `npx spritesh` command doesn't take in account this). In any case the SVGs are automatically copied now in #171 so probably we can close this one, and later check in #171 that the icons are rendered correctly in the mini-website.

Dhaulagiri (approved):
I don't see the conflicting IDs but I'm happy to revisit if it's still a problem with the sprite after it's generated via the build.

didoo:
@Dhaulagiri for future reference: the `paint0_linear` ID is not unique in the sprite:
screenshot_76
screenshot_75

PR #196, opened 2021-09-14
Export script (Node version) - Building (PR)

didoo:
## :pushpin: Summary

Created a "build" script that takes the SVG files exported from Figma via the "sync" script (see #164) and generates a set of "bundles" ready to be published as NPM modules, plus updates the Ember Addon SVG sprite and catalog.json files.

I have also re-organized/updated the documentation (README/CONTRIBUTING) in the project root, and the `flight-icons` and `ember-flight-icons` to make it more in sync with the current status of the project.

_Notice: this solves #171._

### How to review this PR.

Probably, looking at the diff in GitHub can be difficult, there are a lot of changes so it may be difficult to compare the before and after. My suggestion is to checkout the branch, install the dependencies and run the "build" script in your local environment (see "How to Build and Test the Change" below) to see how it works and its output.

### Things to look at:

- the overall content/organization of the folder `flight-icons`
- the main script`fligth-icons/scripts/build.ts` and its sub-modules in the folder `fligth-icons/scripts/build-parts`
- the format of the generated files in the bundles (expecially the optimized SVGs and the SVG sprite)

- Checkout the branch of this PR
- Install the dependencies: `cd flight-icons/scripts && yarn install`
- Create a `.env` file in the `scripts` folder with your Figma API key (see CONTRIBUTING.md in the /flight-icons/ folder)
- Run the "sync" script (optional): `yarn sync`
- Run the " build" script: `yarn build`
- Check the generated files in the `dist` folder

## :+1: Definition of Done

See #171

## 🤔 Things to decide

- Do we want to commit the `dist` folder in the repo? In theory is not needed, it can always be regenerated from the `src` in any moment.
- The final format of the "custom" Consul CSS file is still under discussion, but this does not change the overall script (it's a matter of convention/agreement with the Consul team on how they want the SCSS generated).

Examples: 


Dhaulagiri:
👏  very cool!

Some notes as I try this out:
* I initially went to `flight-icons` and ran `yarn` because I saw a package.json and was surprised I was in the wrong place. I should have read the instructions more closely, but I could make a (weak?) case that we could merge `flight-icons` and `flight-icon-scripts`? Perhaps this setup will make more sense once we look at the release phase though?
* It seems fine to copy the sprite and catalog file into the ember addon, but it could be slightly confusing that we'll have two instances of `catalog.json` in the project. This is another one we could revisit after the release work is done as well.

> Do we want to commit the dist folder in the repo?

This seems like a pattern I see elsewhere that could be helpful for us. For instance if only 1 icon is changing seeing that 1 icon as part of the commit will be helpful in confirming your change is doing what you think it's doing.

amyrlam:
I'll start taking a look locally, but others can feel free to +1. But... if this depends on https://github.com/hashicorp/flight/issues/200, I agree with the comments there, that that's a React approach and not an Ember one. So we may have to massage this, could be in a fast follow.

flight-icons/scripts/build-parts/generateBundleSVG.ts
```diff
@@ -0,0 +1,18 @@
...
+    const distBundleFolderPath = `${config.distFolder}/flight-icons-svg`;
+
+    // create the destination folder
+    await fs.mkdirs(distBundleFolderPath);
+
+    // copy the assets catalog file
+    await fs.copy(`${config.srcFolder}/catalog.json`, `${distBundleFolderPath}/catalog.json`);
+
+    // copy the optimized SVGs to the "dist" target folder (more correctly, its subfolder)
+    await fs.copy(tempSVGFolderPath, `${distBundleFolderPath}/svg/`);
```
Dhaulagiri, line 17: minor nit: having these in the `svg` subfolder feels redundant given the parent folder is already called `flight-icons-svg`

didoo:
> * I initially went to `flight-icons` and ran `yarn` because I saw a package.json and was surprised I was in the wrong place. I should have read the instructions more closely, but I could make a (weak?) case that we could merge `flight-icons` and `flight-icon-scripts`? Perhaps this setup will make more sense once we look at the release phase though?

I can try, that was an option I thought. The reason why I haven't done it yet, is because the `package.json` is then copied (at release time) in the bundles, and I wanted to keep it as clean as possible. I'll see what I can do.

> It seems fine to copy the sprite and catalog file into the ember addon, but it could be slightly confusing that we'll have two instances of catalog.json in the project. This is another one we could revisit after the release work is done as well.
Well, if you can reference the catalog from the `src` folder in Ember, that would be the solution. But in any case we would have two sprites, so...

> This seems like a pattern I see elsewhere that could be helpful for us. For instance if only 1 icon is changing seeing that 1 icon as part of the commit will be helpful in confirming your change is doing what you think it's doing.
That is totally my thought (and that's what I used to do previously). If it's ok, more than happy to commit also the `dist` (unless someone comes up with counter arguments).

re. the `svg` sub-folder is not redundant, on the contrary: so the catalog.json doesn't get mixed up with the SVG files. Separation of concerns! :)

Dhaulagiri:
> if you can reference the catalog from the src folder in Ember, that would be the solution

In theory/ideally we can reference the catalog and the sprite by installing `@hashicorp/flight-icons-sprite` in the ember app

didoo:
> > if you can reference the catalog from the src folder in Ember, that would be the solution
> 
> In theory/ideally we can reference the catalog and the sprite by installing `@hashicorp/flight-icons-sprite` in the ember app

Yes, that's what I thought. In that case, we will automatically "bump the version" (to the same number) whenever we do a new release of the flight-icons package(s) in the Ember addon dependency (`package.json`). Would that work for you?

Dhaulagiri:
> Would that work for you?

Most likely. Can always see how it feels in practice once we get to the release work and adjust if needed.

amyrlam:
i'm having trouble leaving a review w/comment (files changed is so large the page keeps freezing) so will just click the button there. this is my approval comment!

- this looks fine and i ran it locally. we should have ember-flight-icons consume the sprite then, but seems easier to address as a follow-up, which i can queue up now, meeting with didoo tomorrow so we can talk about it then

amyrlam (approved):

PR #199, opened 2021-09-14
Feature: add functionality to toggle rendered icon sizes in demo app

MelSumner:
## :pushpin: Summary

If merged, this PR hooks up the toggle button to the rendered list of icons so that only the chosen icon size renders. 

image

image

Tests are passing: 
image

Examples: 


MelSumner:
@dizzyup @didoo did anything change with the icons themselves to make them a teensy tiny bit smaller?

Issue #93, opened 2021-08-23
Consider adding Percy


amyrlam:
Would allow us to have some sort of test for visual changes re: https://github.com/hashicorp/flight/issues/87

didoo:
+1 on this, we can even test both the full page of https://flight-staging.vercel.app/ and the single icons for better results (depends on how Percy scripting works; I used a custom solution built on top of Puppeteer for this and worked a charm)

amyrlam:
Yeah as a first pass how it would work from my past experience is:

- set up testing
- visit url
- percy snapshot the page of all the icons

amyrlam:
Wanted to add that we could potentially also set this up in `cloud-ui`, which may help us as they decouple structure / their own in-repo design system to migrate to flight icons / new design system. re: https://docs.google.com/document/d/1zd9vvh66uYVCbs90Ncknsp9yyIaQpt35-w_RyfvHWHo/edit?disco=AAAAOsVRXd4

Dhaulagiri:
This would make more sense for us to consider once we solidify our approach to documentation as any setup work we did now would need to be redone for that.

PR #209, opened 2021-09-14
Update docs

amyrlam:
## :pushpin: Summary

- We can update the vercel link if we want to change the name, but it was incorrect
- Clarified README vs. CONTRIBUTING
- We may want to add Markdownlint but it isn't that important, there's a ticket if becomes an issue down the road

Follow-on to https://github.com/hashicorp/flight/pull/196

### DIFF
--- a/flight-icons/README.md
+++ b/flight-icons/README.md
@@ -6,11 +6,12 @@ A package for the icons from `flight`, which can be used in React or Ember apps.
 
 ## Installation and Usage
 
-See [https://flight-hashicorp.vercel.app/](https://flight-hashicorp.vercel.app/) for installation and usage instructions.
+See [https://flight-hashicorp.vercel.app](https://flight-hashicorp.vercel.app) for installation and usage instructions.
 
 ## Contributing
 
 See the [Contributing](CONTRIBUTING.md) guide for details.
+
 ## License
 
 This project is licensed under the [Mozilla Public License 2.0](LICENSE.md).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,11 +3,13 @@
 [![PRs Welcome](https://img.shields.io/badge/PRs-welcome-brightgreen.svg?style=flat-square)](http://makeapullrequest.com)
 
 We welcome contributions from external members of the team! We encourage you to float the idea for your PR first, via a GitHub issue or a post in Slack.
+
 ## @hashicorp/flight-icons
 
-See the [Readme](flight-icons/CONTRIBUTING.md) guide for details on how to use the "sync/build/release" scripts, that export the assets from Figma and save them as multiple bundles in different formats (standalone SVGs, SVG sprite, CSS, etc.).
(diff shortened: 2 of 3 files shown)


flight-icons/README.md
```diff
@@ -6,11 +6,12 @@ A package for the icons from `flight`, which can be used in React or Ember apps.
 
 ## Installation and Usage
 
-See [https://flight-hashicorp.vercel.app/](https://flight-hashicorp.vercel.app/) for installation and usage instructions.
```
Dhaulagiri, line 9: @amyrlam you might have been out when we were discussing this, what is the difference between these two URLs?

amyrlam:
looking at the test failure rn

flight-icons/README.md
amyrlam: we can change it, i've been directing ppl to the link on the RHS of the repo homepage so should decide on one?

https://hashicorp.slack.com/archives/C025N5V4PFZ/p1630612266013300

Dhaulagiri: but they go to the same place, yes? The thinking was that https://flight-hashicorp.vercel.app/ seemed more production-y than the staging url so it would make sense to start using that one by default moving forward.

amyrlam: image

^ updated 

sounds good, i deleted flight-staging as a domain. i'm not sure why vercel gives us the other one automatically yet, but not super impt

ember-flight-icons/tests/acceptance/index-test.js
```diff
@@ -11,6 +11,6 @@ module('Acceptance | icon index', function (hooks) {
     await visit('/');
     // added this because the icons are not rendering as quickly
     await waitFor('.flight-icon', { timeout: 1000 });
-    assert.dom('[data-test-target="icon-grid"] [data-test-icon]').exists();
+    assert.dom('[data-test-target="icon-grid"]').exists();
```
amyrlam, line 14: I think we could test this in Percy instead in the future, cause (I think) technically the `ul` could be there with no icons displaying https://github.com/hashicorp/flight/issues/93

amyrlam: https://github.com/hashicorp/flight/pull/211

amyrlam:
- We went with https://flight-hashicorp.vercel.app
- I updated the docs
- I did Percy separately (https://github.com/hashicorp/flight/pull/211) so this is just a docs tweak

MelSumner (approved):
Looks good!

ember-flight-icons/tests/acceptance/index-test.js
Dhaulagiri: This change more or less makes this test not worth having imo as the point of the test was to confirm the sprite loaded _at all_ and this is just confirming the page has loaded.

amyrlam: i was confused by this test TBH, cause it wasn't that way in the DOM with the icons displaying? i can add it back, but maybe a named hook on the icon itself makes more sense, in addition to the percy check? how about i make a PR to propose what i mean

Issue #200, opened 2021-09-14
Ember addon <SVG> - fill vs style


didoo:
In the new Node pipeline I have moved the "currentColor" declaration from the <SVG> node directly in the <path> nodes (see #196). The reason for this is that now we don't "build" anymore the SVGs via script, we download directly from 
Figma, and Figma of course doesn't provide the `fill` in the <SVG> node. To have that, would have meant to do SVG manipulation, which I decided to avoid because:
- it's more common to have the `currentColor` applied directly on the paths (for example in future we could have multi-color icons, in which case makes even more sense)
- it allows us to have the standalone SVGs with the `currentColor` value directly in the paths (they will be consumed by React via `InlineSvg`)

I have realized that there's a problem with this: the Ember addon uses the `fill` attribute on the <svg> to apply the `color` prop. What I am suggesting here is to change the way we apply this color. 

Currently the prop value is applied as fill:
```
<svg [...] fill="{{this.color}}" />
```
but I suggest to apply it as inline style:
```
<svg [...] style="color: {{this.color}}" />
```

@MelSumner if it's OK for you, I can easily prepare a PR. Otherwise, if you have other suggestions, I'm all ears.

meirish:
Using inline style is disallowed for HCP's Content-Security-Policy (and I think for Vault's as well?), so I don't think `style` will work for those apps.

meirish:
Oops forgot to link https://github.com/jelhan/ember-style-modifier - we could use that in the ember addon to apply styles in a CSP-safe fashion, in the provided component, but I'm not sure if there are other considerations.

MelSumner:
> Using inline style is disallowed for HCP's Content-Security-Policy (and I think for Vault's as well?), so I don't think `style` will work for those apps.

Ember will also throw an error if you use inline styles.

MelSumner:
I can work around this change but if we could have an `isFullColor` property from the Figma API that would be nicer/easier.

didoo:
@MelSumner can you expand on this? I don't follow/understand

MelSumner:
@didoo: 
appending `-color` to the name of the icon is going to prevent users from setting a fill, so I was thinking I could conditionally support the `@color` property by doing this: 
- check to see if the name of the icon (`this.args.name`) includes '-color'
- if it does, don't include the `fill` attribute on the component
- maybe give an error message that the `@color` can't be set on a full-color icon?

[...]

Issue #198, opened 2021-09-14
Update Ember Addon to inline the SVG sprite and not reference it as external resource


didoo:
There is an issue with using the SVG sprite as external resource:

```
<svg>
    <use href="/@hashicorp/ember-flight-icons/icons/sprite.svg#bitbucket-color-24"></use>
</svg>
```
vs instead inlining the SVG sprite in the page and then referencing it via its ID:

```
<svg>
    <use href="#bitbucket-color-24"></use>
</svg>
```

When used in the first way, the "definitions" in the sprite are not considered (see couple of links at the bottom).

This means that the gradients applied to the icons are not rendered in the Ember addon (that's why we have the issue mentioned by @Dhaulagiri in https://github.com/hashicorp/flight/pull/193#issuecomment-918628049. At the beginning I thought it was because of non unique IDs (which was also a problem, but then it was fixed but the issue remained), so I made a quick test (see [test.zip](https://github.com/hashicorp/flight/files/7162012/test.zip); you can serve the folder via `python -m SimpleHTTPServer 8000`) and yes, the issue is with the external reference:

<img width="163" alt="screenshot_77" src="https://user-images.githubusercontent.com/686239/133260347-c0391e79-9c72-4363-86dd-fb50581b2d19.png">

@MelSumner @Dhaulagiri do you think there is a way to fix this in the Ember addon? (embedding the SVG in the page, instead of referencing it as remote resource?) I know we're losing the benefit of caching a resource (btw, in that case probably we should have some cache-busting mechanism in place, for when the sprite is updated, otherwise a user may not see the updates) but I don't know any alternative, I've always used inline SVGs.

The alternative may be to get rid of the gradient at all in Figma but:
- we are playing with other companies logos, not good
- we may want in the future include ourselves icons with gradients

------

Relevant links
- https://fvsch.com/svg-gradient-fill
- https://stackoverflow.com/questions/63898270/linear-gradient-in-svg-symbol-while-using-svg-sprite

MelSumner:
I can try it but we have already been told that caching is not an issue, because most product consumers have to run the entire app locally anyway (or something like that).

didoo:
@MelSumner so if caching is not an issue, let's have the SVG embedded in the page (don't know how to do it in Ember, otherwise I would have opened a PR)

Dhaulagiri:
[...]

Issue #144, opened 2021-09-02
Verify svgo working as expected


amyrlam:
> @johncowen hey, feel free to chime in whenever i would say!

👍 

> icons should not be scaled in new code. but... (1) there may be exceptions in new code. (2) that may not be possible / feasible in all legacy code.
> 
> we've run into situations already where the icon is in line with typography that's 20px high in boundary-ui, or a smaller icon like 11px is being used in learn. so... looking for us to flesh out the guidelines as we're trying to build adoption of the package

aaahhh right gotcha, thanks for the info. In that case my first thought here would be maybe 'should'?

> ..., consumers **should not** stretch

Something like that ^ maybe is more or less what we are saying here? Dunno up to you folks.

> there are 2 packages: https://www.npmjs.com/package/@hashicorp/ember-flight-icons and https://www.npmjs.com/package/@hashicorp/flight-icons. i think you may be able to use flight-icons with the css approach in consul, but i haven't dug into it yet, we could pair next week if you want? (i'll be out fri & tue, extending the long weekend)

Yeah for sure, it's super good for us at the moment 👍 

I had to double look here as I thought I'd misunderstood which README this was for, we are talking about the README for the icons only package here right? Maybe I've misunderstood that this is meant for the all the ember stuff on top also, sorry if so!

> i thought the icons were svgo'd, can you be more specific about what are you seeing? i can file that as a separate bug ticket if can get more detail! also, feel free to file atomic issues on flight at will, helps for tracking

Oh weird, maybe I'm doing something wrong 👀 , will take a look first to check. The thing I noticed was, as Consul inlines the SVGs into CSS using a script, whilst it does this it doesn't actually inline the SVG onto a single line, so when I ran the script on these icons the output was all broken:

```css
background-image: url('stuff:<svg><g></g></svg>')
/*vs*/
background-image: url('stuff:<svg>
  <g>
  </g>
</svg>')
```

My first thought was "oh I need to run svgo on the icons", when I did that it removed all the whitespace from the svg and the CSS output script then worked like:

```css
background-image: url('stuff:<svg><g></g></svg>')
```

I think I remember that you can pass different config to svgo or something, maybe its that, I was just using `svgo -i ./folder`.

_Originally posted by @johncowen in https://github.com/hashicorp/flight/issues/119#issuecomment-911427964_

Dhaulagiri:
We can confirm the correct behavior when we re-write the export. It is very possible this isn't actually working right now.

[...]

Issue #77, opened 2021-08-19
Match number of icons from Figma with number of icons exported


Dhaulagiri:
Given I have N icons in the Figma file
The exporter should have a validation in place to confirm it generates N svgs

didoo:
probably can be tackled in #28

amyrlam:
Can also be cross-checked with https://github.com/hashicorp/flight/pull/211

didoo:
@amyrlam what you mean? there is already a check in place here: https://github.com/hashicorp/flight/blob/73ec29b468e07ffb828a635a8e36633cd38f3cc5/flight-icons/scripts/sync.ts#L58, how Percy would check?

amyrlam:
sounds good, let's mark it done! just discussed IRL

PR #210, opened 2021-09-14
[not for merge] Testing node for pairing tmrw 

amyrlam:
- [ ] See if https://github.com/hashicorp/flight/issues/144 is sorted 
- [ ] #spritegate https://github.com/hashicorp/flight/issues/198 (not certain if this is a 1.0) 
- [ ] https://github.com/hashicorp/flight/issues/200 from DM, sounds like didoo will change the script and not use inline style. confirm at meeting 
- [ ] https://github.com/hashicorp/flight/issues/77 seems like can be closed


flight-icons/scripts/sync.ts
```diff
@@ -37,6 +37,16 @@ import { config } from './config';
 async function sync() {
 
     // remove existing output folder
+    /* 
+        Question: do we need to run this every time, or could we just overwrite icons that have changed? 
+        No strong opinion, just curious of the thought here.
```
didoo, line 42: Icons can also be removed, or renamed, not just added (rare, but it happens). That's why it's easier to just delete the output folder and don't have special logic to handle that.

flight-icons/scripts/sync.ts
```diff
@@ -37,6 +37,16 @@ import { config } from './config';
 async function sync() {
 
     // remove existing output folder
+    /* 
+        Question: do we need to run this every time, or could we just overwrite icons that have changed? 
+        No strong opinion, just curious of the thought here. 
+        I guess the catalog.json has the date, so the corresponding SVGs should also be fresh.
+        
+        - Is a good next step to have ember-flight-icons consume flight-icons? 
+        - I'm not really following flight-icons-svg vs. flight-icons-svg-sprite. Those could also be packages if we want everything atomic?
```
didoo, line 46: Let's discuss together, I can explain the thinking (and what options I see).

didoo (commented):
Added a couple of comments

amyrlam:
this all makes sense from us pairing today, was just testing in public, closing!

PR #218, opened 2021-09-15
Internal: refactored some demo app css

### LINKED ISSUE #203: Internal: tidy up CSS for docs site
The CSS in the docs site is a little bit messy for a couple of reasons: 

1. we intentionally decided to not decide what kind of CSS to do
2. we then decided to use Tailwind on the docs site

So we need to clean up the CSS. 

- [ ] reduce all of the specifications to only use the `.ds-` classes
- [ ] convert non-tailwind css to tailwind css (for consistency, where possible)
- [ ] sort the properties uniformly

MelSumner:
## :pushpin: Summary

If merged, this PR tidies up class names and consolidates CSS in the demo app to use Tailwind classes. Should not materially change what is seen in the browser, but may improve view on smaller viewports in some cases.

Resolves #203

Examples: 


MelSumner:
I think this PR will clean up things I intended to tidy up once the functionality was completed; there might be some more adjustments to make but those should come with conversation first, I think.

didoo (approved):
didn't do a full review line by line, but skimmed through the code and checked the deployed mini website, and all looks good to me. 

nice job! 👌

PR #219, opened 2021-09-15
Bugfix: resolves issue where the data-test-icon attribute had no value

### LINKED ISSUE #31: Set the value of the data-test-icon attribute to the name of the icon
Small thing, but `<svg data-test-icon={{@name}} ...>` might be useful for testing?

MelSumner:
## :pushpin: Summary

If merged, this PR resolves the issue where the `data-test-icon` attribute had no value. It now explicitly sets the value to `{{@name}}`.

The docs have also been updated to reflect this change. 

The `ember-test-selectors` ([an Ember testing addon](https://github.com/simplabs/ember-test-selectors)) has been added to help ensure that `data-test-*` attributes are stripped properly from production builds. 

Resolves #31 

Examples: 

### DIFF
--- a/ember-flight-icons/addon/components/flight-icon.hbs
+++ b/ember-flight-icons/addon/components/flight-icon.hbs
@@ -2,7 +2,7 @@
   class="flight-icon flight-icon-{{@name}} {{if this.display "flight-icon-display-inline"}}"
   ...attributes
   aria-hidden="true"
-  data-test-icon
+  data-test-icon={{this.iconId}}
   fill="{{this.color}}"
   id={{this.iconId}}
   width="{{this.size}}"
--- a/ember-flight-icons/addon/components/flight-icon.hbs
+++ b/ember-flight-icons/addon/components/flight-icon.hbs
@@ -2,7 +2,7 @@
   class="flight-icon flight-icon-{{@name}} {{if this.display "flight-icon-display-inline"}}"
   ...attributes
   aria-hidden="true"
-  data-test-icon={{this.iconId}}
+  data-test-icon={{@name}}
   fill="{{this.color}}"
   id={{this.iconId}}
   width="{{this.size}}"
--- a/ember-flight-icons/package.json
+++ b/ember-flight-icons/package.json
@@ -93,6 +93,7 @@
     "ember-source": "~3.26.1",
     "ember-source-channel-url": "^3.0.0",
     "ember-template-lint": "^3.5.1",
+    "ember-test-selectors": "^6.0.0",
     "ember-truth-helpers": "^3.0.0",
     "ember-try": "^1.4.0",
     "eslint": "^7.27.0",
--- a/ember-flight-icons/package.json
+++ b/ember-flight-icons/package.json
@@ -58,7 +58,8 @@
   },
   "dependencies": {
     "ember-cli-babel": "^7.26.6",
-    "ember-cli-htmlbars": "^5.7.1"
+    "ember-cli-htmlbars": "^5.7.1",
+    "ember-test-selectors": "^6.0.0"
   },
   "devDependencies": {
     "@ember/optional-features": "^2.0.0",
@@ -93,7 +94,6 @@
     "ember-source": "~3.26.1",
     "ember-source-channel-url": "^3.0.0",
     "ember-template-lint": "^3.5.1",
-    "ember-test-selectors": "^6.0.0",
     "ember-truth-helpers": "^3.0.0",
     "ember-try": "^1.4.0",
     "eslint": "^7.27.0",
--- a/ember-flight-icons/tests/dummy/app/templates/engineering.hbs
+++ b/ember-flight-icons/tests/dummy/app/templates/engineering.hbs
@@ -15,6 +15,8 @@
   <p>To install, run
   <CodeBlock @code="yarn add @hashicorp/ember-flight-icons" @language="bash" />
   </p>
+  <p>Note: <kbd class="ds-kbd">ember-test-selectors</kbd> is a dependency added for the author's convenience. 
+  <a href="https://github.com/simplabs/ember-test-selectors" class="ds-a">This Ember Addon</a> strips out all <kbd class="ds-kbd">data-test-*</kbd> attributes for production builds.</p>
 </div>
 
 <div class="mt-8 prose prose-lg text-gray-800 pb-6">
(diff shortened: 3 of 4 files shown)


Dhaulagiri:
My opinion is that it would be more useful to set the value to something more human readable like the icon's file name. Asserting against something like `[data-test-icon="icon-ember15"]` feels unintuitive and also likely to break in weird ways if the dom structure in a consuming app changes.

ember-flight-icons/addon/components/flight-icon.hbs
```diff
@@ -2,7 +2,7 @@
   class="flight-icon flight-icon-{{@name}} {{if this.display "flight-icon-display-inline"}}"
   ...attributes
   aria-hidden="true"
-  data-test-icon
+  data-test-icon={{this.iconId}}
```
didoo, line 5: thought: maybe we can double check wit the teams using this attribute(s) what they expect this value to be. (I for example would expect to have the icon name as value; in my previous company these `data-test` attributes were using in automation testing to select a DOM element or check their existence, so having a random value would not work, but maybe here is different). 

suggestion: we could even discuss if they want/need more than one attribute (in that case expose a `data-test` as object of keys/values)

MelSumner: @dhaulagiri if I change it to name, will it cause issues if it's a test for a page with two of the same icons on it?

Dhaulagiri: It could, but my expectation would be that the developer should have something local to that icon invocation that they can use to scope the selector more narrowly. Or they would be testing a component in isolation that is less likely to have multiple instances of a single icon. 

This is where @didoo's suggestion of taking a look at some existing codebases would be helpful as I'm pretty sure the structure component is already doing something similar.

Either way, I'm pretty sure that having it be based on the id would cause this value to be non-static which seems problematic for test scenarios.

amyrlam:
I don't see https://github.com/hashicorp/flight/issues/31 as a 1.0 showstopper, how did we come to that decision? I would argue that it's better to set the testing hook as needed, not on every icon? It's not common for other components to come with a "hook" out of the box, no?

amyrlam:
From the Slack thread I added `atlas` folks to participate

MelSumner:
> I don't see #31 as a 1.0 showstopper, how did we come to that decision? I would argue that it's better to set the testing hook as needed, not on every icon? It's not common for other components to come with a "hook" out of the box, no?

@amyrlam not a showstopper at all, just working through some of the issues in the repo.

didoo:
> I don't see #31 as a 1.0 showstopper, how did we come to that decision? I would argue that it's better to set the testing hook as needed, not on every icon? It's not common for other components to come with a "hook" out of the box, no?

@amyrlam not a blocker for the 1.0 release, but for the adoption in Atlas, remember we discussed and agreed with @junyper ?

amyrlam:
sounds good, is why I added junyper and thrashr888 as reviewers to see if it will work for their needs, they have the most context on this issue with `atlas` / relates to the slack thread going on rn https://hashicorp.slack.com/archives/C11JCBJTW/p1631731593173900

amyrlam:
@didoo to clarify, https://github.com/hashicorp/flight/pull/220 is the blocker for `atlas` adoption and i added them as reviewers there as well

ember-flight-icons/tests/dummy/app/templates/engineering.hbs
```diff
@@ -27,7 +27,7 @@
     <li class="ds-li"><kbd class="ds-kbd">height</kbd> and <kbd class="ds-kbd">width</kbd>: default size of 16x16 (px)</li>
     <li class="ds-li">(CSS)<kbd class="ds-kbd">class</kbd>: flight-icon, flight-icon-NAME, flight-icon-display-inline</li>
     <li class="ds-li">CSS display: set to <kbd class="ds-kbd">display:inline-block</kbd></li>
-    <li class="ds-li"><kbd class="ds-kbd">data-test-icon</kbd> attribute: for the author's testing convenience</li>
+    <li class="ds-li"><kbd class="ds-kbd">data-test-icon</kbd> attribute: for the author's testing convenience; set to the value of the auto-generated (unique) id.</li>
```
MelSumner, line 30:
```suggestion
    <li class="ds-li"><kbd class="ds-kbd">data-test-icon</kbd> attribute: for the author's testing convenience; set to the value of the <kbd class="ds-kbd">@name</kbd> property.</li>
```

ember-flight-icons/tests/dummy/app/templates/engineering.hbs
```diff
@@ -37,12 +37,12 @@
...
 
   <p>
-  Renders to this:
+  Renders to this (where the ID will be unique each time):
 <CodeBlock @language="markup" id="component-invocation-base-render"
 @code="&lt;svg 
  aria-hidden=&quot;true&quot; 
  class=&quot;flight-icon icon-alert-circle display-inline&quot; 
- data-test-icon
+ data-test-icon=&quot;icon-ember115&quot;
```
MelSumner, line 45:
```suggestion
 data-test-icon=&quot;alert-circle&quot;
```

ember-flight-icons/package.json
```diff
@@ -93,6 +93,7 @@
     "ember-source": "~3.26.1",
     "ember-source-channel-url": "^3.0.0",
     "ember-template-lint": "^3.5.1",
+    "ember-test-selectors": "^6.0.0",
```
Dhaulagiri, line 96: this may be :canofworms: so we can table if so, but in theory we'd want this behavior to ship with the addon so in consuming apps the `data-test-*` is stripped from production builds. The downside is that if the app isn't already doing that we would be introducing that behavior for them which _could?_ have unintended consequences? It definitely shouldn't though.

Dhaulagiri: HCP does have this addon [installed already](https://github.com/hashicorp/cloud-ui/blob/d2f33217d797b58a2c0f1106ec15d6cfc92a76d2/hcp/package.json#L117)

amyrlam (commented):
this seems g2g to me now, but looks like it needs a rebase, due to the percy diff

also happy to discuss this PR at the next design system sync meeting in real time as there have been a lot of comments in various channels on github/slack, along with https://github.com/hashicorp/flight/pull/220

MelSumner:
I've put this on the agenda for our team discussion today. I'll move it to draft until after that discussion.

amyrlam:
trying one rebase to resolve the percy flakiness. if doesn't work will just call it a loss!

amyrlam (approved):
looks good, let's get this in and iterate as needed 

FWIW the percy diff was solved with a rebase on `main`. we should be try a rebase first before approving a flaky diff, to prevent percy churn

PR #220, opened 2021-09-15
Temporary Enhancement: adds a title property to the component

### LINKED ISSUE #29: Add a title argument to the FlightIcon component
I'd like to recommend adding a `title` optional argument.

Note: the last time I checked using `aria-labelledby` along with the `title` element worked consistently across screen readers to provide an accessible label for an inline SVG.

The template would be something like this:

```
<svg 
  aria-hidden={{unless (is-empty @title) "true"}}
  role={{if (is-empty @title) "presentation" "img"}}
  aria-labelledby={{this.titleId}}
  class="flight-icon icon-{{@name}} {{if this.display "display-inline"}}"
  data-test-icon={{@name}}
  fill={{this.fillColor}} 
  height="{{this.size}}"
  id={{this.iconId}}
  viewBox="0 0 {{this.size}} {{this.size}}"
  width="{{this.size}}" 
  xmlns="http://www.w3.org/2000/svg"
>
  {{#if @title}}
    <title id={{this.titleId}}>{{@title}}</title>
  {{/if}}
  <use href='{{this.baseURL}}/ember-flight-icons/icons/sprite.svg#{{@name}}-{{this.size}}'></use>
</svg>
```

With this implementation, developers would get the presentational markup by default, but could easily assign a title when it makes sense to do so.

MelSumner:
## :pushpin: Summary

If merged, this PR:

- adds a `@title` property to the Ember component as a stopgap to help bridge current and planned future use
- adds more tests
- updates engineering documentation

Resolves #29 

- added ability to define a value for `@title` property (e.g., `<FlightIcon @name="zap" @title="meaningful title goes here" />`)
- if `@title` is defined, these things happen: 
  - a `titleId` is automatically generated and added to the title element
  - a `<g>` element with `role="presentation"` is wrapped around the `<use>` element
  - `aria-hidden` is set to `false` on the parent `<svg>` element
  - `aria-labelledby` is added to the parent `<svg>` element and the value is set to `titleId`
  - `role="img"` is added to the parent `<svg>` element

In the future, the design system will hopefully eliminate the need for this additional complexity by creating the components that seem to currently require all of these extra properties/attributes. Users should try to avoid using this if at all possible, and instead think about how to achieve accessibility from the context (e.g., a `<button>` element with `aria-label` defined).

Examples: 


amyrlam:
i actually don't have `atlas` local dev running yet, so i added @junyper as reviewer, as will have the most context (i haven't seen the use case in the browser myself yet). thanks for doing this!

ember-flight-icons/addon/components/flight-icon.hbs
```diff
@@ -1,7 +1,7 @@
 <svg 
   class="flight-icon flight-icon-{{@name}} {{if this.display "flight-icon-display-inline"}}"
   ...attributes
-  aria-hidden="true"
+  aria-hidden="{{if @title "false" "true"}}"
```
junyper, line 4: I think you need to also add `aria-labelledby={{if @title this.titleId}}"` to link the title to the `svg` element, so that it is read by all screen readers. (Unless the support for the `title` element has changed since I last tested). This complexity is why I thought it would be good to include this in the component.

ember-flight-icons/tests/dummy/app/templates/engineering.hbs
```diff
@@ -9,6 +9,25 @@
...
 </div>
 
+<div class="mt-8 prose prose-lg text-gray-800 pb-6">
+  <h2 class="ds-h2" id="accessibility"><a href="#accessibility" class="ds-a" rel="external">&sect;</a> Accessibility</h2>
+  <p>
+  Accessibility (a11y) support for SVGs is inconsistent across browsers and assistive technology. Currently, best practice is to set the <kbd class="ds-kbd">aria-hidden</kbd> attribute to false on the SVG itself.
+  This means that the icon (both the singular icon and the icon component) will need to be used <em class="ds-em">in context</em>.
+  The icons themselves are for presentation purposes only and should never be used on their own.
+  </p>
+  <p>However: As a <em>temporary</em> bridge while we work to provide the accessible components in the design system, we have provided the ability to add a title element to the Ember component by defining a value for the <kbd class="ds-kbd">@title</kbd> property.
```
junyper, line 19: This note seems misleading to me as (I think) we've shown several uses of icons in our products where they are used to replace text content but are not interactive.

junyper (changes requested):

ember-flight-icons/addon/components/flight-icon.hbs
```diff
@@ -1,7 +1,7 @@
 <svg
```
junyper, line 1: Sorry I missed this earlier. You also need to set `role={{if @title "img" "presentation"}}` for the title attribute to work correclty.

MelSumner: This last one will be a blocker; I'm going to put this back into draft mode until the DS team has time to discuss further. 

Thank you for the great feedback @junyper!

ember-flight-icons/tests/dummy/app/templates/engineering.hbs
MelSumner: Thank you for your feedback. We are encouraging folks to work with their designers to move on from this pattern. 👍

ember-flight-icons/addon/components/flight-icon.hbs
didoo: nitpick: since below we have `aria-labelledby={{this.ariaLabelledby}}` would it make sense (and be more consistent) to have a similar function for this value? something like `aria-hidden={{this.ariaHidden}}`. 

(or alternatively convert line 6 to `aria-labelledby={{if @title this.titleId null}}`?)

ember-flight-icons/addon/components/flight-icon.hbs
```diff
@@ -10,5 +12,12 @@
   viewBox="0 0 {{this.size}} {{this.size}}"
   xmlns="http://www.w3.org/2000/svg"
 >
-  <use href='{{this.contextRootURL}}@hashicorp/ember-flight-icons/icons/sprite.svg#{{@name}}-{{this.size}}'></use>
+  {{#if @title}}
+    <title id={{this.titleId}}>{{@title}}</title>
+    <g role="presentation" class="flight-icon-g">
```
didoo, line 17: thought: since this CSS selector is used only for testing (as far as I can see, I didn't see any CSS change) would it make sense to use a selector like `assert.dom('svg > g')` in the test?

ember-flight-icons/addon/components/flight-icon.hbs
```diff
@@ -10,5 +12,12 @@
...
+      <use href='{{this.contextRootURL}}@hashicorp/ember-flight-icons/icons/sprite.svg#{{@name}}-{{this.size}}'></use>
+    </g>
+  {{else}}
+    <use href='{{this.contextRootURL}}@hashicorp/ember-flight-icons/icons/sprite.svg#{{@name}}-{{this.size}}'></use>
+  {{/if}}
```
didoo, line 22: praise: nice! an elegant piece of code (this whole bloc) :)

ember-flight-icons/addon/components/flight-icon.js
```diff
@@ -64,4 +64,53 @@ export default class FlightIconComponent extends Component {
   get size() {
     return this.args.size ?? '16';
   }
+
+  /**
+   * Generates a unique ID for the title, if one exists
```
didoo, line 69:
```suggestion
   * Generates a unique ID for the title
```

ember-flight-icons/addon/components/flight-icon.hbs
MelSumner: yes! I'll improve it.

ember-flight-icons/addon/components/flight-icon.js
MelSumner: added in the commit that updates the test 👍

ember-flight-icons/addon/components/flight-icon.hbs
MelSumner: Thank you ^_^

ember-flight-icons/addon/components/flight-icon.hbs
amyrlam: to clarify, seems like can remove this class, is unused now?

MelSumner:
```suggestion
    <g role="presentation">
```

didoo (approved):

Issue #207, opened 2021-09-14
Internal: jquery test scenario not passing


MelSumner:
jquery is still a test scenario in ember-try. It's currently not passing. We should figure out how to make it happy or remove it from the test scenario list.

Dhaulagiri:
Is there a specific point at which it stopped passing?

Dhaulagiri:
It looks like something with the docs changes has introduced some [flakiness in our tests](https://github.com/hashicorp/flight/pull/208/checks?check_run_id=3603559453)

amyrlam:
I also saw ember-release fail here: https://github.com/hashicorp/flight/pull/210

It should be solved by this https://github.com/hashicorp/flight/pull/209#discussion_r708645146, I may quickly try Percy today (and timebox) cause may be a better test anyways? Please LMK if I am missing something

MelSumner:
I started noticing the flakiness after hooking up the toggle this morning. I might have inadvertently introduced a race condition, will need to track it down more.

amyrlam:
This should be solved as I removed `assert.dom('[data-test-target="icon-grid"] [data-test-icon]').exists();` which wasn't in the DOM https://github.com/hashicorp/flight/pull/209#discussion_r708645146. Removed as a part of the Percy test replacement https://github.com/hashicorp/flight/pull/211. It was that test failing thruout the various scenarios.

Please reopen if you see again though!

Dhaulagiri:
I don't think this is actually solved as we are simply masking the problem by not testing for it now. We haven't actually solved the underlying issue with there being some unpredictable delay in how quickly the sprite gets loaded. Perhaps that's will be a moot point though if we need to change how the sprite is loaded for https://github.com/hashicorp/flight/issues/198

amyrlam:
ok, this relates to https://github.com/hashicorp/flight/pull/209#discussion_r709262671 cause the only test failing was cause of the DOM thing? let me reopen and track that work here

amyrlam:
i'll wait for https://github.com/hashicorp/flight/issues/31 to land

PR #221, opened 2021-09-15
Add back test, with more specificity

amyrlam:
## :pushpin: Summary

Add back test, with more specificity 

It's possible that in CI, the test was waiting for the icons at the top linking to the docs in the rectangular grid, not the full set of icons below, as can't repro locally.

Note: Update this PR if https://github.com/hashicorp/flight/pull/219 lands, maybe with the first data-test-icon name in .ds-icon-frame

## :link: External Links

Resolves: https://github.com/hashicorp/flight/issues/207

### DIFF
--- a/ember-flight-icons/tests/acceptance/index-test.js
+++ b/ember-flight-icons/tests/acceptance/index-test.js
@@ -9,9 +9,11 @@ module('Acceptance | icon index', function (hooks) {
 
   test('visiting / renders a list of icons', async function (assert) {
     await visit('/');
-    // added this because the icons are not rendering as quickly
-    await waitFor('.flight-icon', { timeout: 1000 });
+
+    // added timeout because the icons are not rendering as quickly
+    await waitFor('.ds-icon-frame > .flight-icon', { timeout: 1000 });
+    assert.dom('[data-test-target="icon-grid"] [data-test-icon]').exists();
+
     await percySnapshot('Icons page');
-    assert.ok(true);
   });
 });


MelSumner (approved):
looks good 👍

ember-flight-icons/tests/acceptance/index-test.js
```diff
@@ -9,9 +9,11 @@ module('Acceptance | icon index', function (hooks) {
 
   test('visiting / renders a list of icons', async function (assert) {
     await visit('/');
-    // added this because the icons are not rendering as quickly
-    await waitFor('.flight-icon', { timeout: 1000 });
+
+    // added timeout because the icons are not rendering as quickly
```
Dhaulagiri, line 13: @MelSumner @amyrlam do we have any idea why the icons are not rendering quickly (you can see this when you load the docs)? Is it something specific to how we are rendering them in the docs or is it inherent to the sprite? This makes me concerned that folks using the addon could experience a similar delay which would not be good.

amyrlam: I don't see it loading that slowly in tests. Even running an `ember test --server` with a debugger, it was hard to say. I also ran some of the ember try:once commands locally

Does Vercel have caching out of the box? I could honestly set up Fastly for us in 30-60min, it would just be a straight up config with gzip (I think)

Dhaulagiri: I'll make a separate issue to better capture this

Dhaulagiri: https://github.com/hashicorp/flight/issues/226

Issue #115, opened 2021-08-26
Look into release-it for npm releases


amyrlam:
https://github.com/release-it/release-it

@MelSumner uses it here https://github.com/ember-a11y/ember-a11y-refocus/blob/master/RELEASE.md

MelSumner:
@Dhaulagiri @didoo I think we should integrate this into the Ember addon. It takes care of some of the otherwise manual (and therefore subject to potential error) stuff out of the way.

didoo:
@MelSumner yes, I was looking into Lerna as possible solution to release multiple packages at the same time. I'll add `release-it` to the list of things to look into with @amyrlam in the context of the "release" step of the Node script.

amyrlam:
synced with didoo. i am going to give release-it a spin now, and cristiano will check out lerna

amyrlam:
It seems safe to close this as took a different approach in https://github.com/hashicorp/flight/pull/225. Let's have Melanie, Cristiano, or Brian do the next npm release to test our process.

BTW if I'm incorrectly closing, feel free to reopen the issue!

PR #222, opened 2021-09-16
Update npm docs

amyrlam:
## :pushpin: Summary

This is a slight update to the docs as-is. 

//

I'm still looking at release-it (https://github.com/hashicorp/flight/issues/115), is my first time using it. 

I see some benefit to automating it... but we have to keep in mind that one package could be bumped differently than the other. There are instances where we will only bump `@hashicorp/ember-flight-icons` and not `@hashicorp/flight-icons`. So... I would suggest we keep the releases atomic, which I think relates to the overarching goal of Cristiano's vision (flight-icons-svg could be consumed by ember-flight-icons, a potential future react-flight-icons etc.)

The idea behind the ticket is to get release-it to release both ember-flight-icons  package.json and flight-icons package.json within the "monorepo" we have right now, perhaps with https://github.com/rwjblue/release-it-yarn-workspaces which Cristiano shared.

I like Melanie's suggestion of using release-it and release-it-lerna-changelog https://github.com/hashicorp/flight/issues/115#issuecomment-906539536, which links to https://github.com/ember-a11y/ember-a11y-refocus/blob/master/RELEASE.md ... but keep in mind this isn't a huge automated improvement to what we have as-is in this PR. 

I would argue that we _do_ want to discuss SemVer as a team in a PR, to make sure we agree on what's patch/minor/major, especially if we have to maintain multiple versions of a style-guide. Which is why I kind of prefer having that be a part of code review. 

Can continue to work on release-it schemes tomorrow though, even if it's adding it to both packages as an incremental step forward.

### DIFF
--- a/ember-flight-icons/CONTRIBUTING.md
+++ b/ember-flight-icons/CONTRIBUTING.md
@@ -24,13 +24,13 @@
 
 For more information on using ember-cli, visit [https://ember-cli.com/](https://ember-cli.com/).
 
-## Releasing a new version of the package
+## 🚧 [WIP] Releasing a new npm version of the package
 
 ```bash
 cd ember-flight-icons
 ```
 
-Bump the version number for the `ember-flight/package.json`.
+Bump the version number, per SemVer, for the `ember-flight/package.json`.
 
 After the change is merged to `main`, from the `ember-flight-icons/` directory, run:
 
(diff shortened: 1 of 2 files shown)


MelSumner (approved):
I think we assume that they are subject to change until we hit 1.0, but I don't mind having it explicitly defined. 👍

amyrlam:
Yes, we can use release-it or other, but rn there is some drift in the docs. I would like us to publish a beta version of both addons and we haven't actually tested what we have today also

Issue #188, opened 2021-09-13
Export script (Node version) - Releasing


didoo:
#### Description
This is the last part (the "releasing") of the new Node.js version of the "Figma to Flight" export script.
What it does is to take the bundles generated in the "build" step in different formats, prepare them as NPM packages and release them to the registry.

The Ember Addon instead will be released following the standard process in Ember (but we'll try to find a way to automate also this release).

#### Expected logic
This is an outline of what we intend to implement

```
- clone (or just process, if we don't commit it to the repo)) the `dist` folder that was generated by the "build" step
- bump the version of the main package.json file and all the derived ones [notice: TBD the versioning strategy]
- prepare the "bundles" folders to be released as NPM packages
- release each target package to NPM
  - TBD if we need to check that the use has already done a login and has the right permissions
- cleanup any temporary file we created in the process
```
#### Things to consider/keep in mind

- We have discussed the versioning strategy (see #154) and we have agreed that the version of the packages and the version of the Ember addon will be independent.
- We still have to understand if it's possible to run a `yarn release` (or similar) command via Node script, or needs to be done via CLI (in which case the "release" process may change a little).

#### Tools & Links
- Yarn Workspaces - https://classic.yarnpkg.com/en/docs/workspaces/
- Lerna - https://lerna.js.org/
- Lerna Wizard - https://github.com/webuniverseio/lerna-wizard
- Release-it - https://github.com/release-it/release-it (see #115)
- a couple of examples in which Lerna is used with monorepo+Ember packages: [example 1](https://github.com/dunkinbase/ember-elements) - [example 2](https://github.com/freshworks/nucleus)
- https://github.community/t/multiple-packages-from-the-same-repo-root-level-package-json-needed/17691
- npm-publish-all (package) - https://www.npmjs.com/package/npm-publish-all
- https://medium.com/ah-technology/a-guide-through-the-wild-wild-west-of-setting-up-a-mono-repo-with-typescript-lerna-and-yarn-ed6a1e5467a
- how to install a local node package manager https://github.com/lerna/lerna/issues/51#issuecomment-348256663 (plus thread about how to do a dry-run of `lerna publish`)

#### Tasks
- [x] Pair with @amyrlam and understand what options we have, what is the expected content of the bundles to be ready to be published, and what we can do to automate this process as much as possible
- [x] Check how we can release multiple packages with a single command
- [x] Setup a process that takes the bundles and release them on the registry
- [x] Update the README file with the documentation for the "release" step

[...]

PR #224, opened 2021-09-16
Export script (Node version) - Releasing (PR - Part #1: Flattening folder structure)

didoo:
## :pushpin: Summary

This is the first part of the PR to solve #188. 

In this PR I have flattened the structure of the `flight-icons` folder, removing the `src` and `dist` subfolder and moving everything up in the "main" folder (apart from the scripts). 

Now there is one single `catalog.json` file, the exported "original" SVGs are saved in `svg-original` while the processed SVGs are stored in `svg`.

In this way the folder is more inline with the `ember-flight-icons` folder structure, and there's no need to copy stuff around when we publish it to NPM (see https://github.com/hashicorp/flight/issues/188#issuecomment-920923890).


amyrlam (approved):

amyrlam:
this looks good - thank you - couldn't leave a comment as part of PR approval process cause the PR is so large and freezing the browse 😅

PR #225, opened 2021-09-16
Export script (Node version) - Releasing (PR - Part #2: Node script to publish)

didoo:
## :pushpin: Summary

This is the second part of the PR to solve #188. 

The **important** :) one.

In this PR I have added entries in the `package.json` files of `flight-icons` and `ember-flight-icons`, to run scripts that bump the version in the `package.json` file (interactively) and the publish the package to the NPM repository.

We have decided to use this simple package ([version-bump-prompt](https://github.com/JS-DevTools/version-bump-prompt)) to do interactive bumping (the user can easily select which semver bump they want to do), which is very similar to what [release-it](https://github.com/release-it/release-it) does from what I understood from @amyrlam and @MelSumner.

This should now allow us to release packages in production, from now on.

### DIFF
--- a/ember-flight-icons/package.json
+++ b/ember-flight-icons/package.json
@@ -54,7 +54,10 @@
     "test": "npm-run-all lint test:*",
     "test:ember": "ember test",
     "test:ember-compatibility": "ember try:each",
-    "test:ember:percy": "percy exec ember test"
+    "test:ember:percy": "percy exec ember test",
+    "release:bump": "./node_modules/.bin/bump package.json",
+    "release:publish": "yarn publish --new-version $npm_package_version",
+    "release": "yarn release:bump && yarn release:publish"
   },
   "dependencies": {
     "ember-cli-babel": "^7.26.6",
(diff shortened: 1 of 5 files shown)


ember-flight-icons/package.json
```diff
@@ -54,7 +54,10 @@
     "test": "npm-run-all lint test:*",
     "test:ember": "ember test",
     "test:ember-compatibility": "ember try:each",
-    "test:ember:percy": "percy exec ember test"
+    "test:ember:percy": "percy exec ember test",
```
amyrlam, line 57: (just an aside) how do we feel about trailing comma in prettier? i kinda like it cause makes diffs easier to read, but no strong opinion. i don't even think we're prettier-ing the overall repo in CI beyond `ember-flight-icons` but don't quote me on that

amyrlam (approved):
lgtm, let's give it a try!

ember-flight-icons/package.json
didoo: don't think you can add extra commas in JSON files (I have to double check but I am pretty sure)

PR #230, opened 2021-09-16
Add “services” icons to the “structure-mappings” file

didoo:
## :pushpin: Summary

I have added the “services” icons to the “structure-mappings” file.

@dizzyup I have added only the "mono" version, should I add also the `-color` ones too? (in that case the name is the same, also for the old ones?)


didoo:
> ## 📌 Summary
> I have added the “services” icons to the “structure-mappings” file.
> 
> @dizzyup I have added only the "mono" version, should I add also the `-color` ones too? (in that case the name is the same, also for the old ones?)

Never mind, found the answer: https://www.figma.com/file/B7yd8loYS7nA2TiQcJ8UtV/Structure-1.9?node-id=1152%3A7099 :)

dizzyup (approved):
Looks good to me!

Issue #213, opened 2021-09-15
Investigate linting in GitHub Actions CI


amyrlam:
Not urgent but just noting as something nice to have

- Is linting running as expected? Didn't seem to be failing this week on weird .hbs formatting 
- Add Prettier? Is here https://github.com/hashicorp/flight/pull/221/checks?check_run_id=3615666388 but check if beyond ember-flight-icons child folder, if desired
- Look into pre-commit hooks, Husky? There was a comment about this

PR #232, opened 2021-09-17
Add no-trailing-spaces rule for cleaner .hbs file diffs

amyrlam:
## :pushpin: Summary

Add no-trailing-spaces rule
Ref: https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-trailing-spaces.md
Prevent noisy diffs, noticed on #229

Resolves some of: https://github.com/hashicorp/flight/issues/213

PR #233, opened 2021-09-17
Add prettier to hbs

amyrlam:
## :pushpin: Summary

- Set up prettier for .hbs
  - Ref: https://dev.to/jelhan/format-glimmer-templates-with-prettier-ipa
- Run yarn run lint:hbs to prettify code

I made a PR to boundary today and noticed I didn't even have Prettier on in my editor. With this PR, Prettier will be checked by CI and diffs will have less noise.

Resolves: https://github.com/hashicorp/flight/issues/213 (Seems good enough to me, pre-commit hooks are more of a nice to have since can run everything with the package.json scripts / extensions autosave, etc. But could make another ticket if we want that.)

Examples: 


didoo (commented):
Preamble: I'm 100% for using Prettier on JS/TS/JSX files, even CSS probably.

That said, I don't think we should use Prettier for HTML (and so consequently HBS).
The way it formats the code it's counter-intuitive for an HTML developer, and I have seen a lot of changes in the code in this PR that don't make sense from the perspective of readability of HTML.

My point of view, but I wait to see other people's reviews.

MelSumner (approved):
Thank you for adding this!

amyrlam:
@didoo Can you give an example of what kind of changes you've seen? Wondering if a role we can toggle on/off with ember-template-lint 

I guess the benefit of Prettier is you can write your code in any way. I could add a pre-commit hook to autofix so users can write their code if it's a real blocker, I think we can do that in the yarn scripts without installing Husky or similar

didoo:
> @didoo Can you give an example of what kind of changes you've seen? 

screenshot_88

amyrlam:
@didoo Thanks for sharing, I think most of those things are from line length. Can try to adjust the line length for just .hbs later today and see what you think...

amyrlam:
@didoo or others - check the latest and let me know what you think 

I think this is better than the status quo with no linting, my motivation in doing this is to make PR's easier to review.

Would be easier to do in a follow-in PR but:

- I can make a pre-commit hook (maybe an optional one in the package.json scripts?) so that the developer can write .hbs in their own way and then it's formatted on commit. 
- Can also add instructions for how to set up autoformat in VSCode to CONTRIBUTING.
- I think we can add some sort of file also, that will have VSCode load up all the appropriate extensions for this `flight` repo, if the developer opts-in.

didoo:
@amyrlam I'm still not sold. now the documentation files are all mixed up.

I don't want to be a pita (I'm on a 5 on a [Cap Watkins figs scale](https://capwatkins.com/blog/the-sliding-scale-of-giving-a-fuck)) but I want to make sure that everyone is onboard with this (I'm talking about using prettier for HTML and HBS files). From what I see @MelSumner is OK with it, if also @Dhaulagiri is ok, let's do it.

Dhaulagiri:
I definitely agree with @didoo  that some of these changes are not what I would have expected or preferred personally, but it's a low risk change that we can always disable if it proves too unpleasant.

I think the one thing I would be explicitly 👎  on is that this is also impacting the code examples which I don't think we would want to do:

### Before
Screen Shot 2021-09-20 at 08 37 13


### After
_More indentation of the props and the closing tag_
Screen Shot 2021-09-20 at 08 37 01

amyrlam:
ok makes sense... hm let me look at the non-prettier options out there (is not my priority or anything). some of the indentation rn seemed off, adding whitespace changes to diffs

MelSumner:
Maybe we could have it mostly turned off but with useful things like "remove whitespace at the end of a line" turned on? WDYT?

amyrlam:
Fixed trailing whitespace here: https://github.com/hashicorp/flight/pull/232 ... will tweak this one and then see what y'all think, happy to close if undesired!

ember-flight-icons/tests/dummy/app/templates/engineering.hbs
```diff
@@ -2,82 +2,139 @@
...
+  id=&quot;icon-ember115&quot;
+  width=&quot;16&quot;
+  height=&quot;16&quot;
+  viewBox=&quot;0 0 16 16&quot;
+  xmlns=&quot;http://www.w3.org/2000/svg&quot;>
   &lt;use href=&quot;/@hashicorp/ember-flight-icons/icons/sprite.svg#alert-circle-16&quot;>&lt;/use>
-&lt;/svg>" />
+&lt;/svg>"
+    />
+    {{! prettier-ignore-end }}
```
amyrlam, line 128: Added prettier ignore's to every CodeBlock. FWIW CodeBlock has kinda unusual indentation as it is... L126 has to be "un-indented" to display properly on the page. 

I think this is an improvement to what we have and FWIW `boundary-ui` uses `ember-template-lint-plugin-prettier` and `cloud-ui` plans to

Can do a pre-commit hook to `yarn run lint:fix` maybe with https://pre-commit.ci/ to make it a better DX

But happy to close if y'all still disagree

amyrlam:
made some edits, ready for another review, lmk what you think

ember-flight-icons/tests/dummy/app/templates/engineering.hbs
amyrlam: talked with @didoo and @MelSumner 

let's use a pre-push hook instead, that will lint on push

melanie shared https://github.com/okonet/lint-staged which i will look into!

PR #229, opened 2021-09-16
Export script (Node version) - Releasing (PR - Part #3: Documentation)

didoo:
## :pushpin: Summary

This is the last part of the PR to solve #188. 

In this PR I have:
- updated the CONTRIBUTE documentation for both the `flight-icons` and `ember-flight-icons`, reorganizing some content and adding specific instructions to follow for the "release" of npm packages.
- made a small change to a path in the documentation for developers in the mini-website

PR #236, opened 2021-09-17
Split the “bump” command from the actual “publish” action

didoo:
## :pushpin: Summary

I have split the “bump” command from the actual “npm publish” action. In this way they can be run independently, and we can have a specific PR on the update of the version (per conversation in: https://github.com/hashicorp/flight/pull/229#discussion_r710986676)


amyrlam (approved):
👍 

we've been saying approver can merge, so i'll merge it in as no unresolved comments, CI is green

PR #243, opened 2021-09-20
Tweak instructions

amyrlam:
## :pushpin: Summary

Following up from the meeting, what do you think about this change? Not super important, just a suggestion

I added the console.log back, but wasn't sure if there was a reason it was commented out?


MelSumner (approved):
👍

heatherlarsen (approved):
lgtm 👍

amyrlam:
i'll merge in and if didoo has changes tmrw will fix in a follow-on!

didoo:
@amyrlam I'm ok with the changes (only the console.log, are you sure? it adds a lot of noise :) in any case we can comment it later, so all good for me)

amyrlam:
good pt, how about https://github.com/hashicorp/flight/pull/259? tagged you there as reviewer

PR #244, opened 2021-09-20
Bugfix: resolves Safari issue

### LINKED ISSUE #228: Docs site icon search field misbehaves on Safari
As you’ll see in the video, as I type the input is erased and replaced with the last typed character (and the cursor position is reset).

https://user-images.githubusercontent.com/34030/133685405-ad1882c3-93c2-4662-9a8c-f50299978ced.mp4

MelSumner:
## :pushpin: Summary

If merged, this PR resolves issue where a Safari bug was preventing correct use of the live-filter on the icons. 

Resolves #228 

Examples: 


amyrlam (approved):
weirdest bug haha. nice one!

PR #245, opened 2021-09-20
Inline the SVG sprite in the HTML page when using "ember-flight-icon" addon

Dhaulagiri:
## :pushpin: Summary

Resolves https://github.com/hashicorp/flight/issues/198

This attempts to resolve https://github.com/hashicorp/flight/issues/198 by inlining the SVG within these constraints:
* We can't load the SVG from a CDN
* We can't load the SVG via an inline script

* In the vercel app, look at the `bitbucket-color-*` icons and see the gradient and compare to the non-gradient in production docs

Examples: 


didoo:
Two things:

1. I can confirm now the issue with the gradients is fixed:
screenshot_96

2. the SVG sprite injected as is will add extra space in the page (it occupies a 300x150px block). 
screenshot_95
we need to make it hidden (but not via `display: none`), eg. setting `position: absolute; width: 0; height: 0; overflow: hidden;` for the <svg> not via inline style, since is Ember, but as class I suppose :)

ember-flight-icons/index.js
```diff
@@ -2,4 +2,32 @@
...
+
+              document.body.appendChild(svg);
+            });
+
+            // BLOCKER: How to get at the sprite _here_
+            req.open('GET', '/@hashicorp/ember-flight-icons/icons/sprite.svg');
+
+            req.send();
+          })();
+        </script>
```
johncowen, line 29: 👋  not quite sure what the overall thinking is here, but if you are going forwards with some sort of inline script, Consul (and Vault/CloudUI also probably and I would imagine other products in the future) doesn't allow inline scripts for security reasons. So if this is the direction you are going then this will need to be an external script if possible.

didoo:
@Dhaulagiri (and @amyrlam @MelSumner) see my latest commit in this repo, and at the diff, and let me know what you think. 
it works on my local machine, in the dummy app (micro-website), now it's a matter of testing in other applications (expecially on cloud-ui as you suggested in the Slack thread). so if you need we can do a new package release from this branch, with already these changes in it.

ember-flight-icons/index.js
Dhaulagiri: Thanks for the heads up on this. We're going to try something slightly different to attempt to avoid this.

Dhaulagiri:
@didoo I did some commit cleanup to remove my initial WIP work and also made a slight tweak to your change to keep the file we generate in `/public` instead of putting it in `addon/components`. I'm not 100% sure if public is "correct" but I'm fairly confident having it live in the other folder is fairly non-idiomatic and likely to cause some amount of confusion down the line. This still seems to work locally testing the docs site and linked against waypoint. If @amyrlam or @MelSumner are 👍  we could ship this and cut another beta release to test more concretely in cloud-ui. If it doesn't work we can always revert this PR.

Dhaulagiri:
🤔  it seems potentially instructive/a bad sign that the percy snapshots _are not_ loading the icons at all

ember-flight-icons/index.js
```diff
@@ -1,5 +1,13 @@
...
+const flightIconSprite = require('./public/icons/flight-icon-sprite');
+
 module.exports = {
   name: require('./package').name,
+
+  contentFor(type) {
+    if (type === 'body-footer') {
+      return flightIconSprite;
+    }
+  },
```
amyrlam, line 12: nice - thanks, i couldn't remember what we had done here

i'm taking a look now

amyrlam:
It's VERY strange that the snapshot for https://github.com/hashicorp/flight/blob/main/ember-flight-icons/tests/dummy/app/templates/percy-test.hbs is broken with this change...even though it's just calling a straight up `<FlightIcon>`. I don't think we should ship this until that's fixed. 

I emailed Percy support, also noticed that the emoji on /percy-test is broken (but that was on `main` as well).

Unfortunately I'm about to step into a 2hr meeting from onboarding. Will update you here by EOD, even if it is "no update"

Dhaulagiri:
> It's VERY strange that the snapshot for https://github.com/hashicorp/flight/blob/main/ember-flight-icons/tests/dummy/app/templates/percy-test.hbs is broken with this change

it makes sense if the index is broken the other page would be broken too. the sprite is never getting loaded (for some reason) so `<FlightIcon>` in either place wouldn't work.

amyrlam:
moved to draft while we wait to hear back from percy

MelSumner (approved):
Ok looks good, although weird that the preview deploy works but Percy doesn't. 

Let's try it and see how it goes!

MelSumner:
> Oh I get what you mean now - my bad! - I will look more carefully next time before commenting (I didn't think the each of the icon grid was displaying a `FlightIcon`, but given the toggle, that doesn't make sense)

Can you clarify what doesn't make sense?

amyrlam:
@MelSumner that comment is stale, sorry for the confusion. i was debugging why the percy browser doesn't show the inline svg. i mistakenly thought that the icon search grid was mapping over the raw .svg's, not `FlightIcon`. i should have looked at the code first

in https://github.com/hashicorp/flight/pull/261, brian did a test where he is injecting a div with a class into the DOM. it shows up in chrome, firefox and edge, even `ember test --server`, but not percy

i sent percy a support ticket because we don't know what browser they are using. somewhere in their docs it says "a browser we trust"

doesn't seem like a high priority now, but i can ping friends at percy if we don't hear back from them within a week or so

PR #250, opened 2021-09-20
Fix link

amyrlam:
## :pushpin: Summary

Small change to fix link including whitespace

Could potentially be prevented in the future w/something like https://github.com/hashicorp/flight/pull/233

## :camera_flash: Screenshots

WAS:
image

IS:
image

Also found and updated:
image


MelSumner (approved):
Ahh yes, one of the current prettier drawbacks. I hope we can figure this out because I'm always trying to remember to manually fix it before committing. Anyway, thanks for fixing!

amyrlam:
FWIW Prettier isn't implemented on .hbs yet, so this is a different error

PR #251, opened 2021-09-21
[WIP] Fix pre styling: looking for feedback, then can make change thruout

amyrlam:
## :pushpin: Summary

Change `<pre>` styling which is currently markdown backticks

What do you think of this change, used what's on https://tailwindcss.com/

Note: I'm not sure the context behind the `prose` class, it's bringing in a lot of cascading classes. If it were me I would remove `prose prose-lg` and have the typography be atomic. Then could just set the styling for `pre`. But for now, I'm just trying to make an incremental change to improve the look and feel of the page

If you like this change, can swap out all the backticks

## :camera_flash: Screenshots

WAS:
image

IS:
image

### DIFF
--- a/ember-flight-icons/tests/dummy/app/styles/app.css
+++ b/ember-flight-icons/tests/dummy/app/styles/app.css
@@ -33,6 +33,7 @@ html {
   font-family: var(--font-body);
   font-size: var(--font-size-default);
 }
+
 body.ds-body {
   line-height: 1.2;
   @apply
@@ -43,6 +44,7 @@ body.ds-body {
   min-h-screen
   p-0;
 }
+
 header.ds-header {
   @apply
   bg-brand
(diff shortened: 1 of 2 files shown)


amyrlam:
Just noticed this styling on the other page:
image

How about calling that `pre` instead of `.ds-kbd`. It's not really a `<kbd>` anyways?

MelSumner:
so `<pre>` gets different styling- I generally use `<kbd>` if it's inline like this. WDYT?

amyrlam:
it may be better to trend toward semantic HTML so is self-doumenting? seems like kbd is more for like, press <kbd>esc</kbd> to close the modal

i always thought of code as multi-line and pre as in-line, but maybe i'm wrong re: https://stackoverflow.com/questions/5943607/inline-pre-formatting. a class could differentiate multi-line and in-line, but using code or pre would be more clear?

amyrlam:
i'll open a new branch proposing what i think will make sense from a semantic html first way

PR #253, opened 2021-09-21
Fix SVG sprite generation for Ember addon

didoo:
## :pushpin: Summary

 If merged, this PR will:
- remove the `<XML>` and `<DOCTYPE>` extra tags from the generated SVG sprite
- add an extra `style` attribute with value `width: 0; height: 0; visibility: hidden; position: absolute` to avoid the embedded SVG to be visible in the page (this will fix also an issue in #245)
- add an extra `aria-hidden="true"` attribute

(essentially, I have changed the generation to do **exactly** what it did before the conversion of the script to Node)


amyrlam (approved):

amyrlam:
per the meeting merging this in

PR #266, opened 2021-09-23
disable percy

Dhaulagiri:
## :pushpin: Summary

Until we can figure out why the changes coming in for https://github.com/hashicorp/flight/pull/245 are causing Percy to not work we should stop running those checks


amyrlam:
i'm going to comment it out instead, will squash it up

amyrlam (approved):
i went ahead and committed directly to the branch, so that percy no longer appears in the PR CI list. since this is innocuous, i'll go ahead and merge

PR #267, opened 2021-09-23
Update CONTRIBUTING.md

amyrlam:
## :pushpin: Summary

we can rewrite the percy-test snapshot later


MelSumner (approved):
Let's ship it!

didoo (commented):
For me it was much cleaner before. Now I have to read (and think) to understand. Before I could just scan quickly the text, and straight to the point.

amyrlam:
i'll drop the second commit and merge in as-is

PR #271, opened 2021-09-23
Ember addon import `flight-icons` as package

### LINKED ISSUE #272: Have sprite give more descriptive href's
IS:

```
<symbol id="activity-16" viewBox="0 0 16 16"><path fill-rule="evenodd" d="M6.016 1a.75.75 0 01.698.521l3.306 10.33 1.47-4.341A.75.75 0 0112.2 7H15a.75.75 0 010 1.5h-2.262l-2.028 5.99a.75.75 0 01-1.424-.011L5.952 4.06 4.504 8.008A.75.75 0 013.8 8.5H1A.75.75 0 111 7h2.276l2.02-5.508c.11-.301.4-.499.72-.492z" clip-rule="evenodd"/></symbol>
```

and 

image

SHOULD BE:

```
<symbol id="flight-activity-16" viewBox="0 0 16 16"><path fill-rule="evenodd" d="M6.016 1a.75.75 0 01.698.521l3.306 10.33 1.47-4.341A.75.75 0 0112.2 7H15a.75.75 0 010 1.5h-2.262l-2.028 5.99a.75.75 0 01-1.424-.011L5.952 4.06 4.504 8.008A.75.75 0 013.8 8.5H1A.75.75 0 111 7h2.276l2.02-5.508c.11-.301.4-.499.72-.492z" clip-rule="evenodd"/></symbol>
```

Proposal:
- Have a more specific namespaced id in sprite generation
  - So that the href won't collide, in the off chance there is same name
  - Arguably more clear to the dev where it's coming from if was `href=#flight-activity-16` or some other name

Came up in pairing with @DingoEatingFuzz so wanted to float the idea

didoo:
## :pushpin: Summary

This PR changes the way in which `ember-flight-icons` are consuming the SVG sprite. Now the sprite is imported as npm module as dependency in `package.json`. 

In doing so I have also fixed #272.

Notice: I have released the following packages on the public npm registry in order to use them in the `ember-flight-icon` dependencies:
- `@hashicorp/flight-icons/0.0.11-rc`
- `@hashicorp/flight-icons/0.0.12-rc`
- `@hashicorp/ember-flight-icons/0.0.5-rc`
- `@hashicorp/ember-flight-icons/0.0.6-rc`

Please @Dhaulagiri  or @amyrlam can you do a last check in the products using `@hashicorp/ember-flight-icons/0.0.6-rc` and see if everything works fine now?

When ready and approved (and we checked it works fine in the product apps), I think we are ready for a 1.0.0 official release (we will have to update the documentation accordingly, but I'll open a new specific PR for that).


ember-flight-icons/index.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-const flightIconSprite = require('./public/icons/flight-icon-sprite');
+const flightIconSprite = require('@hashicorp/flight-icons/svg-sprite/svg-sprite-module');
```
amyrlam, line 3: this is awesome! originally wanted to do this in the `use` (have ember-flight-icons consume the sprite from flight-icons), but couldn't figure it out in ember / couldn't use an external CDN link.

my one comment is...
- i don't think the sprite is necessary in `@hashicorp/flight-icons` rn. so arguably, some folks might complain it's increasing the size
  - 0.0.10-rc = 754 kB 
  - 0.0.11-rc = 1.26 MB
- could the sprite just be in `@hashicorp/ember-flight-icons` or was there a reason you didn't want that?
- i think the catalog coming from `@hashicorp/flight-icons` is great!

***

yes it works in cloud-ui since brian's fix: 
- see https://github.com/hashicorp/cloud-ui/pull/1322
- at root, run `yarn && yarn start`, log in with github, see external-link icons on the first page 
- last i knew the repo may have oddities with vercel, where the deploy must be requested, don't know much about it yet

didoo: >don't think the sprite is necessary in @hashicorp/flight-icons rn. so arguably, some folks might complain it's increasing the size. the real "problem" is that the initial HTML is much bigger now, but for what I know about how modern browsers work today, this should not be an issue.

it's just in a package, and sits in the `node_modules` folder. if it's not used in the code, it will not increase the final bundle size

> could the sprite just be in @hashicorp/ember-flight-icons or was there a reason you didn't want that?

we had two options to import the "sprite module":
- directly as local file in the `ember-flight-icons` (initial version we did yesterday)
- via npm package.json (current version)

@Dhaulagiri preferred the second, so we don't have to do a new release every time we release a new icon (using the `^1.0.0` annotation in the requirement for the package). I am still not convinced of this approach, I would prefer to release anyway a new version of the `ember-flight-icons` every time, because otherwise we have to require to the developers to manually re-run the install of the package every time we release a new version. In any case, with the current implementation, we can update the package.json in `ember-flight-icons` automatically, if we want, so it's just a matter of deciding what is the best option for us.

> i think the catalog coming from @hashicorp/flight-icons is great!

yep! :)

MelSumner (approved):
Nice improvements, I like it!

Dhaulagiri (approved):
👏

PR #280, opened 2021-09-24
Use more semantic html in /engineering

amyrlam:
## :pushpin: Summary

Clean up .css file for readability

Rename `kbd` to `code`, to follow semantic HTML …
- Ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd

Fix missing space
Also:
- Lint
- Add missing `code` callouts

***

revamp of https://github.com/hashicorp/flight/pull/251/files. i realize it seems minor, and that `ember-prism` brings in a LOT of global styles, but i think it's important that we trend toward semantic HTML

if you review commit-by-commit, may be easier to follow. pls lmk what you think

### DIFF
--- a/ember-flight-icons/tests/dummy/app/templates/engineering.hbs
+++ b/ember-flight-icons/tests/dummy/app/templates/engineering.hbs
@@ -24,7 +24,8 @@
   <p>However: As a
     <em>temporary</em>
     bridge while we work to provide the accessible components in the design system, we have provided the ability to add
-    a title element to the Ember component by defining a value for the <code class="ds-code">@title</code>
+    a title element to the Ember component by defining a value for the
+    <code class="ds-code">@title</code>
     property. This is a temporary measure and we strongly encourage UI engineering teams to work with their designers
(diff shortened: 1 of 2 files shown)


ember-flight-icons/tests/dummy/app/templates/engineering.hbs
```diff
@@ -198,7 +199,7 @@
     <i>inline-block</i>
     to
     <i>block</i>, set
-    <kbd class="ds-kbd">@isInlineBlock</kbd>
+    <code class="ds-code">@isInlineBlock</code>
```
MelSumner, line 202: I don't know if we can use `<code>` by itself (I think it allows us to, but I don't know if we _should_) so let me check this first; otherwise seems good.

MelSumner (approved):
I can't find the reference I was referring to earlier, so I say we ship it. I'll test on a screen reader later and make sure that everything is working properly. 👍

amyrlam:
sounds good, let me know, happy to give the page a test with VoiceOver as we iterate as well

PR #286, opened 2021-10-01
Add Percy back, works now

amyrlam:
I have a support thread going with Percy. Not sure what changed but the svg's appear on Percy now. So let's add this back. I emailed them to ask why.

Please check out the Percy at the bottom and approve it!

Reverts hashicorp/flight#266


amyrlam:
once i dropped the commit dropping the change to the .hbs /percy-test, this no longer worked

Issue #283, opened 2021-09-28
Flight Icon Request: Additional icons for Marketing


dizzyup:
There's a few icons that the marketing web team need to make a full switch to Flight -- 

**Social icons**
- Facebook
- Twitter
- LinkedIn
- YouTube
- Codepen
- Twitch

I'd propose adding these in the full Services style, ie. proper brand assets with mono and color treatments.

**Additional icons**
- Handshake
- Load Balancer
- Keychain
- Mainframe

These are straightforward enough but will need drawing -- also I'm not looking forward to drawing the handshake.

<img width="416" alt="missing-icons@2x" src="https://user-images.githubusercontent.com/717867/135065371-7b9471e4-607c-4ad1-a7ee-e60b95af3189.png">

**Adjustments**
- List
 
The current `list` icon is very similar to the `menu` icon. I'm going to propose editing the `list` icon to add 'bullets' to the left edge of the glyph to better differentiate and make it a more obvious 'list'.

**Metadata**
A few icons were requested but were actually already in Flight, we'll add some metadata in Figma to make these easier to find

- `learn` -> `college, hat`
- `swap-horizontal` -> `arrows, sctm`

rayelder:
@dizzyup I've create the 10 marketing icons. By end of day, I'll create a single Figma branch for all 10. Does that sound okay?

rayelder:
@dizzyup All icons are approved. We decided to use official social icons. The [Add marketing icons](https://www.figma.com/file/MYiw4kiVpunIMMw0sBkE1t/branch/REuSuhh3SbeCNNrIGt2s5D/%E2%9C%8F%EF%B8%8F-Flight-Development?node-id=708%3A1151) branch is up-to-date.

dizzyup:
Thanks @rayelder! Lovely work.

I've merged the changes in and published an update to the Flight library in Figma. Will work with the team to get packages released here also 👍

Dhaulagiri:
Release `@hashicorp/flight-icons-1.1.0` that includes these new icons

PR #288, opened 2021-10-05
sync/build new marketing icons

Dhaulagiri:
## :pushpin: Summary

Addresses https://github.com/hashicorp/flight/issues/283

~~Spot check new/updated icons in the preview app~~ Does not actually work until we release a new package

Look at the icons in the files changed tab

Examples: 


didoo:
@amyrlam any news/updates on #286 / #287 ? we have a few icons that changed path but is kinda unexpected (it can be just Figma) so having a visual regression test before/after may help to do the comparison

Dhaulagiri:
> so having a visual regression test before/after may help to do the comparison

A realization I had doing this PR is that preview app now only  shows what is in the actual published package (the ember addon relies on `@hashicorp/flight-icons` vs previously having the sprite manually copied in), so they can't be used to see changes like this which is a suboptimal unintended side effect of that change I think 🤔

Dhaulagiri:
Although the intention of the Percy tests was to exercise the behavior of the ember component, not necessarily the individual icons themselves

amyrlam:
yeah, not sure. i think it was fine how it was before where this PR would trigger percy, but no preference. it doesn't even work rn.

as for percy, no update. cristiano - added you to the chain with their support. just pinged a friend there to see if can help as the support time has been very long rn 
- i didn't finish debugging but can use https://github.com/hashicorp/flight/pull/261 or the other PR's again, there were some weird things happening... could def spend more time on it.

didoo (approved):

PR #290, opened 2021-10-06
Create dependabot.yml

Dhaulagiri:
## :pushpin: Summary

I had thought this happened automagically but that is only for security updates. This will create version update PRs every Monday for us to review. 

The main value of this is that any updates to `flight-icons` will propagate through to `ember-flight-icons` without us needing to do it manually. Perhaps we should set the schedule to _daily_ though so that we don't have to wait until Monday for new release to show up?

I'd like to get feedback on whether or not we think this is actually a good idea for `flight-icons` given we have no automated tests to confirm that the package updates aren't breaking something.

## :link: External Links

[Docs](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates)

No way to test this locally that I'm aware of

Examples: 

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "npm" 
+    directory: "flight-icons/" 
+    versioning-strategy: increase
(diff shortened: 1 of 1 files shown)


amyrlam (approved):
👍 good with me

i'll revisit the percy fix today

PR #289, opened 2021-10-06
release flight-icons 1.1.0

Dhaulagiri:
## :pushpin: Summary
closes #283

PR #291, opened 2021-10-06
Bump flight-icons in ember-flight-icons to add icons to viewer

amyrlam:
## :pushpin: Summary

bumping flight-icons package so that new icons like `facebook` are visible on the website. normally this would trigger a percy change, i can work on that ticket today

builds off of https://github.com/hashicorp/flight/pull/288 and https://github.com/hashicorp/flight/pull/289


amyrlam:
if we are following semver, i'm thinking any change, even a package.json bump, would mean this is a patch bump to ember-flight-icons e.g. `1.0.1`?  i can make that change as part of this PR if you agree

amyrlam:
ok, that makes sense. 

yeah maybe a good compromise would be that changes to either package add to a CHANGELOG as part of the PR... so that it's clear when the version is bumped later on? isn't super important, just thinking aloud to get on the same page

Issue #284, opened 2021-09-28
Change ember-test-selectors to peerDependency


amyrlam:
Does flight-icons ship with its own version of `ember-test-selectors`? Maybe a `peerDependency` would be better here to allow apps to specify their own version.

_Originally posted by @meirish in https://github.com/hashicorp/cloud-ui/pull/1428#discussion_r717623327_

amyrlam:
- Amy can research more on the why and report back https://code-trotter.com/web/dependencies-vs-devdependencies-vs-peerdependencies/#peerdependencies

`cloud-ui` has ^4.0.0 and ^4.0.1 in the mono https://github.com/hashicorp/cloud-ui/search?q=ember-test-selectors

`flight` has ^6.0.0 https://github.com/hashicorp/flight/blob/58a02a88832aaf109996d1fea07c90850816a472/ember-flight-icons/package.json#L66

PR #306, opened 2021-10-07
Change ember-test-selectors to peerDependency

amyrlam:
## :pushpin: Summary

Will allow consuming apps to use a different version

Resolves: https://github.com/hashicorp/flight/issues/284

In my mind, this change would also bump `@hashicorp/ember-flight-icons` to a 1.1.0 so wanted to be sure we're on the same SemVer page.

//

This change seemed fine to me because we need ember-test-selectors to be under `dependencies` instead of `devDependencies`.

For context, there was an internal [Slack thread about testing strategy](https://hashicorp.slack.com/archives/C11JCBJTW/p1631731593173900) but the context is: we made a decision to include `data-test-` in the FlightIcon component. I personally was of the opinion that we can leave the testing up to the consuming app. (If we went that route, `ember-test-selectors` could be under `devDependencies`.) But I think what we have now is great, and should move forward with this PR as-is if you agree.

- [ ] Update https://github.com/hashicorp/cloud-ui/pull/1428 if this is merged
- [ ] Publish version bump to npm if this is merged

Examples: 

### DIFF
--- a/ember-flight-icons/package.json
+++ b/ember-flight-icons/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@hashicorp/ember-flight-icons",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "description": "The Ember addon for the HashiCorp Flight SVG icon set",
   "keywords": [
     "ember-addon",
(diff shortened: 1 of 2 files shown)


Dhaulagiri:
@meirish this was something you had noted initially, do you have a good frame of reference we should keep in mind in these situations?

MelSumner:
One of the ways peerDependencies is useful is that it makes sure that the version(s) that is compatible with our addon is also in the HC apps.

amyrlam:
@Dhaulagiri Did you mean to close this instead of merge? Just curious why? Seems like the addon is merged in cloud-ui though

Dhaulagiri:
@amyrlam  I had to do a bit of a double take but this did get merged and released

amyrlam:
I totally see it now, thanks for clarifying, that was confusing in the UI!

Issue #318, opened 2021-10-13
Flight Icon Request: Services - EKS and EC2


dizzyup:
Needed to support a new selection interface in HCP Consul:

![screen_shot_2021-10-11_at_9 31 12_am](https://user-images.githubusercontent.com/717867/137109140-1cbabded-331b-4360-8d01-766b34f62f42.png)

Planning on adding these at the end of the week.

dizzyup:
Added in Figma - pending review and release process cc/ @heatherlarsen

Dhaulagiri:
@dizzyup  do we know who requested these icons?

heatherlarsen:
@Dhaulagiri Emma Luo for HCP Consul

heatherlarsen:
Thanks @Dhaulagiri! Emma has been notified via Slack :)

PR #319, opened 2021-10-18
sync & build

Dhaulagiri:
## :pushpin: Summary

* [x] In preparation for new icons ran a `yarn sync && yarn build`. Expected no results, but got some anyhow in 9ac4e4c759593a876d1a40c9727fe9ec79e70f78  🤔 
* [x] `yarn sync && yarn build` to pickup new icons for #318


Dhaulagiri:
@heatherlarsen this is what I was referring to where I wouldn't have expected any changes to appear if I did a rebuild of the icons. Any idea if these changes mean anything or where they might have come from?

Dhaulagiri:
`bitbucket-color` also seems to have lost its depth 🤔

didoo:
> @Dhaulagiri I don't unfortunately. Looking at the history in Figma, the last meaningful entry was done by Hector "adding icons and metadata to support marketing". After that there's a series of autosaves but none of them have a distinct visual difference that I can see and Figma won't let me inspect the layers when I'm in history mode 🤦‍♀️
> 
> @didoo do you have any idea why/how this would have happened?

It's Figma that changed the generated output, you can see it here (no actual change to the icons): 
screenshot_217

(essentially, they're adding the `component-id` to the linear gradient ID/definition; probably to solve the issue many people had when creating SVG sprites, of having the same definition ID for different gradients).

> Whatever the changes are it seems to have changed how the icons render 🤔

Re. the fact that the icon is not "visible", when I test it locally for me works perfectly:
screenshot_219
and I can see all the linear gradients with their correct IDs.

Not sure why in Chrome the diff is not correct (I suspect a bug in Chrome itself, maybe caused by the fact that the ID is potentially not valid according to this: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/id). In fact in Safari it renders correctly:

screenshot_220

PR #328, opened 2021-11-03
ember flight icon liquid size

### LINKED ISSUE #327: Add "liquid" size to the `ember-flight-icon` addon
In the new Badge component the icon size for the `type=small` variant is `12px`. 

screenshot_297

At the moment our `ember-flight-icon` addon doesn't support this size. We need to introduce a "liquid" size that adapts to the parent container, so we can support also alternative sizes.

_Notice: this is a blocker for https://github.com/hashicorp/design-system-components/pull/1_

/cc @hashicorp/design-systems

didoo:
## :pushpin: Summary

If merged, this PR will fix #327.

I have:
- added support for the “stretched” prop in the “FlightIcon” Ember addon
- updated the “percy-test” page
- updated the "engineering" documentation
- updated the unit tests

At the beginning I thought to use the `size` parameter, with a `stretch` value, but then I realized we use it to extrapolate the asset name, so it would not work (how do we know in that case which one of the two assets in two different sizes to use?).

/cc @hashicorp/design-systems

### DIFF
--- a/ember-flight-icons/addon/components/flight-icon.js
+++ b/ember-flight-icons/addon/components/flight-icon.js
@@ -44,14 +44,16 @@ export default class FlightIconComponent extends Component {
   }
 
   /**
-   * Determines the icon's `display` property
-   *
-   * @param display
-   * @returns the value of `isInlineBlock` if set
-   * @default true
+   * Get a class to apply to the icon based on the display argument.
+   * @method FlightIcon#displayClass
+   * @return {string} The css class to apply to the SVG.
    */
-  get display() {
-    return this.args.isInlineBlock ?? true;
+  get displayClass() {
+    if (this.args.isInlineBlock && !this.args.stretched) {
+      return 'flight-icon-display-inline';
+    } else {
+      return null;
+    }
   }
 
   /**
@@ -63,6 +65,18 @@ export default class FlightIconComponent extends Component {
    */
   get size() {
     return this.args.size ?? '16';
+  }
+
+  /**
+   * Get the SVG width/height depending if the icon is stretched or not
+   * @method FlightIcon#svgSize
+   * @return {object} The width/height to apply to the SVG.
+   */
+  get svgSize() {
+    return {
+      width: this.args.stretched ? '100%' : this.size,
+      height: this.args.stretched ? '100%' : this.height,
+    };
   }
 
   /**
--- a/ember-flight-icons/addon/components/flight-icon.js
+++ b/ember-flight-icons/addon/components/flight-icon.js
@@ -49,7 +49,8 @@ export default class FlightIconComponent extends Component {
    * @return {string} The css class to apply to the SVG.
    */
   get displayClass() {
-    if (this.args.isInlineBlock && !this.args.stretched) {
+    const isInlineBlock = this.args.isInlineBlock ?? true;
+    if (isInlineBlock && !this.args.stretched) {
       return 'flight-icon-display-inline';
     } else {
       return null;
--- a/ember-flight-icons/tests/dummy/app/templates/percy-test.hbs
+++ b/ember-flight-icons/tests/dummy/app/templates/percy-test.hbs
@@ -22,7 +22,7 @@
     <div class="w-3 h-3">
       <FlightIcon @name="bug" @size="16" @stretched="true" />
     </div>
-    <div class="w-8 w-8">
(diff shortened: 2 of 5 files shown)


ember-flight-icons/addon/components/flight-icon.js
```diff
@@ -44,14 +44,16 @@ export default class FlightIconComponent extends Component {
...
-   * @param display
-   * @returns the value of `isInlineBlock` if set
-   * @default true
+   * Get a class to apply to the icon based on the display argument.
+   * @method FlightIcon#displayClass
+   * @return {string} The css class to apply to the SVG.
    */
-  get display() {
-    return this.args.isInlineBlock ?? true;
+  get displayClass() {
```
Dhaulagiri, line 51: The previous logic above defaulted to `true` which meant that by default `flight-icon-display-inline` was added, but with this change that is no longer the case which seems like it would be a breaking change. Is that intentional?

ember-flight-icons/tests/dummy/app/templates/percy-test.hbs
```diff
@@ -15,6 +15,19 @@
...
 </div>
 <br />
+
+<div>
+  <p>Icon with custom sizes (12/32):</p>
+  <div class="flex gap-0.5">
+    <div class="w-3 h-3">
+      <FlightIcon @name="bug" @size="16" @stretched="true" />
+    </div>
+    <div class="w-8 w-8">
```
Dhaulagiri, line 25:
```suggestion
    <div class="w-8 h-8">
```

ember-flight-icons/addon/components/flight-icon.js
didoo: yes, nice catch! I have fixed the qunit tests (some were broken, looking for a class that was not there) and added a new one to cover this case. check my new implementation of the `displayClass` method, see if it's OK (but I am open to more elegant solutions, if you have one)

Dhaulagiri: This seems ok to me. I'm a little fuzzy on why we don't want to allow inline-block in liquid mode but i'm super rusty here so will defer to your expertise.

Dhaulagiri (approved):

didoo:
@MelSumner do you mind do a second pass and, if it's OK, approve too? I prefer a double review on this change.

ember-flight-icons/addon/components/flight-icon.js
didoo: with `display: block`:
screenshot_309

with `display: inline-block`:
screenshot_310

(the pink box is the container)

didoo:
@MelSumner is it ok to merge now?

For future reference, we had a meeting with Melanie and we agreed:
- it's OK to have the classes computed via JS instead of inline in the template, because 
  - this is a more resilient approach for the future
  - when there will be much more complex components, that may need even more (UI) logic to control the class names
  - this approach raises a little bit the "local" complexity, but overall reduces the "systemic" complexity (we have a common pattern adopted for all the components, instead of case by case, so the cognitive load for the maintainers to move from the code of one component to another will be much lower)
- it's ok _for now_ to use "layout containers" (`div.hbs-badge__icon` and `div.hbs-badge__text`) to control the internal layout of the components
  - this "separation of concerns" between _containers_ and _content_ worked extremely well for me in the past
  - we avoid mix&match of class names in the content elements
  - in any case we have set up a meeting in 3/4 weeks in which we will review this decision

/cc @hashicorp/design-systems

MelSumner (approved):
I don't really understand this, but I don't object so I am not going to block it either. Let's ship it and see how it goes.

didoo:
@MelSumner I am not sure how you can have the Badge working, with an icon size of 12px, without this change (unless you style directly the SVG in the Badge, in which case this for me is an issue, we should not).

didoo:
@MelSumner by design this icon should be 12px in size (as the container), instead is 16px (as per SVG width/height without the `stretched` option we have introduce here). Makes sense?

screenshot_319

MelSumner:
> @MelSumner by design this icon should be 12px in size (as the container), instead is 16px (as per SVG width/height without the `stretched` option we have introduce here). Makes sense?
> 
> screenshot_319

yup makes sense- I figured we'd refactor the badge a little after this is released. 👍

PR #332, opened 2021-11-05
Update issue template for new icons

didoo:
I have noticed that designers asking for new icons are not attaching screenshots to the issue or linking to the relevant Figma files.
These are useful information for our team's designers to have context of how the icons are used, possible use cases, etc. (also the screenshots are helpful in the future to understand which icon we were referring to).

We want to make this more evident, so I have added a couple of entries to the list, and removed the previous one (which probably was too generic and seemed optional).


heatherlarsen (approved):
lgtm!

PR #336, opened 2021-11-24
fix bug in flight-icon.js

didoo:
## :pushpin: Summary

I have fixed an error in the `svgSize` method, where the `height` value was returned only if the `stretched` parameter was `true` (and so the `height` attribute was not assigned to the SVG element).

I have also bumped the `ember-flight-icon` addon version to `1.2.1` so if approved can be released straightaway.

### DIFF
--- a/ember-flight-icons/tests/integration/components/flight-icon-test.js
+++ b/ember-flight-icons/tests/integration/components/flight-icon-test.js
@@ -24,13 +24,17 @@ module('Integration | Component | flight-icon', function (hooks) {
     await render(hbs`<FlightIcon @name="activity" />`);
     assert
       .dom('svg.flight-icon.flight-icon-activity')
-      .hasStyle({ height: '16px', width: '16px' });
+      .hasStyle({ height: '16px', width: '16px' })
+      .hasAttribute('width', '16')
+      .hasAttribute('height', '16');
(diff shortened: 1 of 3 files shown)


MelSumner (approved):
Approving so this can move forward, but I am not 100% sure about this syntax. Is there a test that could be added?

didoo:
>  but I am not 100% sure about this syntax

can you expand on this?

> Is there a test that could be added?

you're right! I've added (updated) the tests, check if they're OK for you

ember-flight-icons/tests/integration/components/flight-icon-test.js
```diff
@@ -24,13 +24,17 @@ module('Integration | Component | flight-icon', function (hooks) {
     await render(hbs`<FlightIcon @name="activity" />`);
     assert
       .dom('svg.flight-icon.flight-icon-activity')
-      .hasStyle({ height: '16px', width: '16px' });
+      .hasStyle({ height: '16px', width: '16px' })
```
amyrlam, line 27: Filed an issue for this diff re: tooling tweak, not impt though https://github.com/hashicorp/flight/issues/337

amyrlam (approved):
:shipit:

PR #338, opened 2021-12-02
simplify how we inject sprite for tests

Dhaulagiri:
## :pushpin: Summary

Use existing test-body-footer instead of creating a custom one


Dhaulagiri:
Note: this does not work because of [the way percy](https://github.com/percy/percy-ember/blob/66d01330dd24fa074ba7bc583c1bd345b0b7530c/addon-test-support/%40percy/ember/index.js#L33) relies on the contents being inside of `#ember-testing`.

PR #341, opened 2021-12-07
Add "product" icons

didoo:
## :pushpin: Summary

In this PR I have
- added “Products” to the list of frames to parse in Figma (it's a new frame added by @dizzyup)
- synced the SVG icons with latest version of the [Figma file ](https://www.figma.com/file/TLnoT5AYQfy3tZ0H68BgOr/Flight-Icons?node-id=350%3A2)
  - new "product" icons added
  - renamed existing `vault` icon to `bank-vault` to avoid conflict with "vault" product
  - regenerated the SVG icons in output
  - bumped version 2.0.0 (breaking because of the renaming)

screenshot_485


dizzyup (approved):

MelSumner (approved):
Good catch on the whitespaces in the ID attribute. Do we have a linter? (question, non-blocker).

didoo:
> Good catch on the whitespaces in the ID attribute. Do we have a linter? (question, non-blocker).

That ID is generated by Figma, is the name of the layer, there's not much we can do. What we do in output is to strip out all the non necessary IDs.

didoo:
@MelSumner @dizzyup thanks for the prompt review 🙏

(another couple on your way in a minute)

PR #342, opened 2021-12-07
Bump ember-flight-icon

didoo:
## :pushpin: Summary

In this PR I have:
- bumped the `flight-icons` dependency for `ember-flight-icons` to use the latest version (2.0.0) - see #341 
- updated `ember-flight-icons` package version to 1.2.2

### DIFF
--- a/ember-flight-icons/package.json
+++ b/ember-flight-icons/package.json
@@ -60,7 +60,7 @@
     "postrelease": "TAG_VERSION=\"@hashicorp/ember-flight-icons/${npm_package_version}\" && git tag $TAG_VERSION && git push origin $TAG_VERSION"
   },
   "dependencies": {
-    "@hashicorp/flight-icons": "^1.2.0",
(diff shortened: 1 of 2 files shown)


ember-flight-icons/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hashicorp/ember-flight-icons",
-  "version": "1.2.1",
+  "version": "1.2.2",
```
MelSumner, line 3: Shouldn't this bump a major because flight-icons did?

didoo: mm, I am not sure :) I remember we discussed these cases, but I don't remember what was the decision for cases like this. in any case, better to be on the safe side, you're right. I'll push a change.

didoo: @MelSumner pushed v `2.0.0`

MelSumner (approved):
🚢

PR #355, opened 2021-12-22
remove unused iconHbsCode function

Dhaulagiri:
## :pushpin: Summary

This was added in https://github.com/hashicorp/flight/pull/128 but as far as I can tell is not actually used anywhere


amyrlam (approved):
🗑️

amyrlam:
I glossed over the Percy's. Almost seems like it deleted 2 icons in the Percy. Need to take a closer look

PR #354, opened 2021-12-22
update design to include keywords and ember/react names

Dhaulagiri:
## :pushpin: Summary

Closes #350 
Closes #351

### Before

Screen Shot 2021-12-
22 at 12 11 22

### After

Screen Shot 2021-12-22 at 12 38 02

PR #356, opened 2021-12-22
Bump package, adds "hcp" icon

amyrlam:
## :pushpin: Summary

Bump package, adds `hcp` icon

Can add dependabot for just this package as a safeguard for us as a follow-on

Same ubuntu-latest error as https://github.com/hashicorp/flight/pull/354#issuecomment-999810812, can ignore for now

IS: +2 icons `hcp`
image

***

### DIFF
--- a/ember-flight-icons/package.json
+++ b/ember-flight-icons/package.json
@@ -60,7 +60,7 @@
     "postrelease": "TAG_VERSION=\"@hashicorp/ember-flight-icons/${npm_package_version}\" && git tag $TAG_VERSION && git push origin $TAG_VERSION"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


amyrlam:
Filed https://github.com/hashicorp/flight/issues/357 cause `hcp-color` looks the same as `hcp`, but I can take a look at Figma. Not super impt just noting

Dhaulagiri (approved):

heatherlarsen:
@amyrlam `hcp` and `hashicorp` icons both use the standard hashicorp brand color for the `-color` version which is black, so this is the expected result.

amyrlam:
@heatherlarsen oh thanks, that makes sense, my bad!

PR #358, opened 2021-12-22
bind search queries to url for share-ability

### LINKED ISSUE #238: Flight Icons Website: Add search criteria to query params in URL
At the moment is not possible to share one particular icon (or search) with someone else using the micro-website page. This is actually very useful, and will become a very common use case/need. 

I don't know how much is complex to do in the existing Ember app, but would be great if we could update the `window.history` path while the user types, and add the search text in the URL as query parameter. 

In this case, one could simply copy&paste the URL, and share it (or attach it as link somewhere).

Dhaulagiri:
## :pushpin: Summary

Closes #238 

* Migrated search logic to a controller to allow for binding to query params more easily
* Used ember-concurrency to simplify debounce behavior 
* Added minimal state for when no results are returned
* It would probably be a good idea to add some tests for this?

There are most definitely room for improvements as my octane ember knowledge is pretty limited.

## :camera_flash: Screenshots

query

***

### DIFF
--- a/ember-flight-icons/tests/dummy/app/components/ds-icon-grid.hbs
+++ b/ember-flight-icons/tests/dummy/app/components/ds-icon-grid.hbs
@@ -58,5 +58,7 @@
         <span>{{meta.description}}</span>
       </p>
     </li>
+  {{else}}
+    No icons found
   {{/each}}
 </ul>
--- a/ember-flight-icons/tests/dummy/app/components/ds-icon-grid.hbs
+++ b/ember-flight-icons/tests/dummy/app/components/ds-icon-grid.hbs
@@ -15,8 +15,8 @@
         class="ds-input-search"
         placeholder="e.g. arrow"
         autofocus={{false}}
-        value={{this.search}}
-        {{on "input" (perform this.args.searchIcons value="target.value")}}
+        value={{@searchQuery}}
+        {{on "input" (perform @searchIcons value="target.value")}}
       />
 
     </div>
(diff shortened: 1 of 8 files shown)


ember-flight-icons/tests/dummy/app/components/ds-icon-grid.hbs
```diff
@@ -58,5 +58,7 @@
         <span>{{meta.description}}</span>
       </p>
     </li>
+  {{else}}
+    No icons found
```
amyrlam, line 62: 💯

amyrlam:
I'll take a stab at adding a test in a new commit, and see what you think. Can drop the commit if undesired!

Dhaulagiri:
@amyrlam go for it!

amyrlam (approved):
🔍  +1

sorry for the delay, got sidetracked on something for my trip. made a basic test, g2g if good for you. def feel free to push to the new year also!

amyrlam:
* rebased to fix percy

ember-flight-icons/tests/dummy/app/components/ds-icon-grid.hbs
Dhaulagiri: added a test for this case too now

ember-flight-icons/tests/dummy/app/controllers/index.js
```diff
@@ -0,0 +1,32 @@
...
+      });
+    } else {
+      return icons;
+    }
+  }
+
+  @restartableTask *searchIcons(query) {
+    yield timeout(DEBOUNCE_MS);
+
+    this.query = query;
```
Dhaulagiri, line 29: i did this to make sure the actual query param gets updated based on the search string, but it feels like this shouldn't be necessary if the value is properly bound between the query param and the search input like i was expecting 🤷

didoo:
@Dhaulagiri 👏👏👏

PR #361, opened 2022-01-03
split docs site out of addon's dummy app

Dhaulagiri:
## :pushpin: Summary

This PR splits the documentation for flight out from inside the ember addon into a separate ember app. 

### Benefits
The main benefit of this change is that it more clearly decouples the ember addon which is strictly for consumption in ember apps from the documentation which is for the project as a whole.

### Downsides
The main downside I could think of is that this creates another location we'll need to update dependencies when we release new versions of flight/ember-flight-icons, however we could mitigate this with dependabot. 

Also, depending on how the overall design system matures this documentation may move entirely so making this incremental move within this repo may not be worth the effort, so I'm not opposed to opinions to close this PR.

_I tried to have the commits relatively atomic to tell the story of this checklist_

* [x] Create new docs site under `website`
* [x] `git mv` as many files as possible from the dummy app to maintain history
* [x] Move npm packages from dummy app to website
* [x] setup ci for website
* [x] update docs
* [x] Decide if we want to maintain a dummy app deploy/percy testing on the ember addon?
* [ ] Once we/upon merge we'll need to 
  * [ ] update our vercel config to build out of the `website` directory and not `ember-flight-icons`
  * [x] Update percy config

***


amyrlam (changes requested):
Moving it seems fine to me, but I've requested changes on both Percy snapshots. Doing that here to be clear that as-is, this is breaking some functionality that we need. I can take a closer look at why (1) some icons are missing and (2) the /percy-test page is blank.

I added Melanie as a reviewer in case she has thoughts, since we briefly discussed this at standup!

Dhaulagiri:
@amyrlam I think percy has now figured out that the snapshots are ok since there is no baseline to compare them against given this is technically a brand new app/screenshots, but it's definitely possible I'm off base here

amyrlam (approved):

amyrlam:
Wanted to run it locally, LGTM now

I thought 2.0.1 @hashicorp/flight-icons was adding an icon (`hcp`, see https://github.com/hashicorp/flight/commit/ba1d96385661d817b0d74ead47fbc5e8ec0b5486) and I made https://github.com/hashicorp/flight/pull/362, but there's no Percy change ... and `hcp` exists here. So now I'm confused. But I would suggest we merge this as-is to see how Dependabot works if that's cool?

Dhaulagiri:
> I thought 2.0.1 @hashicorp/flight-icons was adding an icon (hcp, see ba1d963) and I made #362, but there's no Percy change ... and hcp exists here

I think I understand what you were looking for but this makes sense because specifying `^2.0.0` of `@hashicorp/flight-icons` does actually install `2.0.1`.

amyrlam:
@Dhaulagiri yes makes sense, I should have done a `yarn why @hashicorp/flight-icons` but was having a new year senior moment 🤦‍♀️ thx for the help!

website/app/router.js
```diff
@@ -0,0 +1,13 @@
...
+
+export default class Router extends EmberRouter {
+  location = config.locationType;
+  rootURL = config.rootURL;
+}
+
+Router.map(function () {
+  this.route('engineering');
+  this.route('design');
+  this.route('percy-test');
```
didoo, line 12: I am not 100% convinced this should live here, at this point. Since this is actually a test page, maybe (this yes!) should live under the "dummy/test" app? it can be a simple page, with zero styling.
Another VRT test page could be a dynamic page that can be used to test every icon (but in this case the cost of Percy would skyrocket).
Not a blocker, just some ideas in case you want to discuss it.

Dhaulagiri: I agree and I struggled with this question as well. Initially I left `percy-test` behind because it _is_ a test for the ember addon itself, but because it has so much tailwind styling and I removed tailwind from `ember-flight-icons` I ended up migrating it over to the website. I could put it back but it seems like it needs _some_ styling since at least some of the tests rely on testing color behavior/etc.

didoo: @Dhaulagiri if you move it back, I can simplify the page and have it styled without tailwind, it's not complex at all (I'll do it in this PR).

Dhaulagiri: I should be able to get it done this afternoon. One nuance we'll have to sort out with this then is configuring multiple percy/vercel setups to deploy/test both `ember-flight-icons` and `website`.

Dhaulagiri: ✅

didoo: @Dhaulagiri FYI just pushed changes to the "percy-test" pages in the dummy app, in which I have removed every custom class or Tailwind class from the page, and used just inline styles.

Dhaulagiri:
BpGWitbFZflfSUYuZ9

PR #362, opened 2022-01-03
Bump package

amyrlam:
## :pushpin: Summary

Tweak to https://github.com/hashicorp/flight/pull/361, curious to see the Percy re: # of icons

***

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -29,7 +29,7 @@
     "@glimmer/component": "^1.0.4",
     "@glimmer/tracking": "^1.0.4",
     "@hashicorp/ember-flight-icons": "^2.0.0",
-    "@hashicorp/flight-icons": "^2.0.0",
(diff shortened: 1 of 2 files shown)


amyrlam:
I would suggest we see if Dependabot works and not merge this

didoo:
@amyrlam and @Dhaulagiri to understand what to do with this PR

amyrlam:
We can close this PR. The PR as-is was bumping the package, which did not make sense, as the yarn.lock had the latest package, even though the package.json had the previous package (due the the `^`). 

Maybe down the road, we'll run into an issue where we forgot to update one of our packages like https://github.com/hashicorp/flight/blob/main/website/package.json#L31-L32, and can see Dependabot working on this repo. If you disagree, I can file an issue, just let me know

PR #363, opened 2022-01-06
update year in footer

Dhaulagiri:


amyrlam (approved):
:shipit: 

Let's make it dynamic on the real website, I saw a funny thing about it on the internet recently: https://updateyourfooter.com/

Dhaulagiri:
😭  indeed! I know I've solved this in a better way in past jobs but one-offs are always so quick

PR #364, opened 2022-01-06
restore embroider tests

Dhaulagiri:
## :pushpin: Summary

These were disabled because they failed due to issues with code in the dummy app. #361 moved the docs out of the dummy app so these work again now

***


amyrlam (approved):
Good for me if no objections @MelSumner. I remember them failing before, but can't remember why exactly.

Issue #365, opened 2022-01-08
Fix Prettier on /website .hbs


amyrlam:
Is not working, see https://github.com/hashicorp/flight/blob/main/website/app/templates/index.hbs, which has single quotes

May be helpful to look into https://github.com/hashicorp/flight/pull/360 first and wrangle the monorepo package.json's first

amyrlam:
update: keep this closed!

PR #368, opened 2022-01-10
fix prettier on website

Dhaulagiri:
## :pushpin: Summary

Fixes https://github.com/hashicorp/flight/issues/365

***


didoo (approved):

website/.prettierrc.js
```diff
@@ -2,4 +2,13 @@
 
 module.exports = {
   singleQuote: true,
+  overrides: [
+    {
+      files: '*.hbs',
+      options: {
+        singleQuote: false,
```
amyrlam, line 9: This isn't being followed in the .hbs below, something is odd (I looked at this briefly on Friday and then was end of the day). I'll reopen the issue so don't forget

website/app/components/ds-icon-grid.hbs
```diff
@@ -32,10 +32,10 @@
       {{on "click" this.updateVisibleIconSize}}
     >
       <span class="ds-toggle-button-slider">
-        <span class="ds-toggle-16 {{if (eq this.visibleIconSize "16") "ds-toggle-16-active"}}" aria-hidden="true">
+        <span class="ds-toggle-16 {{if (eq this.visibleIconSize '16') 'ds-toggle-16-active'}}" aria-hidden="true">
           16px
         </span>
-        <span class="ds-toggle-24 {{if (eq this.visibleIconSize "24") "ds-toggle-24-active"}}" aria-hidden="true">
+        <span class="ds-toggle-24 {{if (eq this.visibleIconSize '24') 'ds-toggle-24-active'}}" aria-hidden="true">
```
amyrlam, line 38: Not sure why some are single quote and some double quote

website/app/templates/design.hbs
```diff
@@ -73,7 +73,7 @@
           </ul>
         </p>
       </div>
-      {{!-- TODO: fix image sizing --}}
+      {{! TODO: fix image sizing }}
```
amyrlam, line 76: This is kinda weird cause I think this is the default if you command - slash, but no big deal

website/app/components/ds-icon-grid.hbs
amyrlam: you're totally right. i assumed it would escape the quote

website/app/templates/design.hbs
didoo: I had the same issue when adding a comment in HBS, and the linter complained. not sure which one of the two is the right one

amyrlam: filed an issue, haven't tried it yet but may be a better way of using prettier in .hbs: https://github.com/hashicorp/flight/issues/371

PR #367, opened 2022-01-08
[not for merge] Curlies in ember-prism

amyrlam:
Repro for https://github.com/shipshapecode/ember-prism/issues/436 FYI @didoo

PR #375, opened 2022-01-12
Update CodeBlock syntax

amyrlam:
## :pushpin: Summary

Allows us to write code samples as they are, improving code readability

## :link: External Links

Follow-on to ember-prism example https://github.com/hashicorp/flight/pull/367

***

### DIFF
--- a/website/app/templates/engineering.hbs
+++ b/website/app/templates/engineering.hbs
@@ -224,12 +224,14 @@
     <code class="ds-code">@stretched</code>
     attribute:
     {{! prettier-ignore-start }}
+    {{! template-lint-disable no-triple-curlies }}
     <CodeBlock
       @language="markup"
-      @code='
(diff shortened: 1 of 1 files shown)


didoo (approved):
OK for me

MelSumner (approved):
This makes me uneasy but if you and @didoo are okay with it then I have no blockers.

didoo:
@MelSumner I agree is not ideal, but of the possible different ways to do it using `ember-prism` this is the least worst (fully escaping the strings is a no-no for me). 

A few days ago I have looked into possible other syntax highlighters, and I could not find one that could support snippets of code directly in the handlebars code. If you have alternatives happy to discuss.

amyrlam:
What makes you uneasy?

We can explore the code in .js option (or snippet of some sort) https://github.com/hashicorp/design-system-components/issues/128, but if we want an "edit on GitHub" button on the website, the code block won't be a part of the same file. 

Feel free to add more detail to that issue or file another one if a better solution?

Issue #258, opened 2021-09-21
Add lint-staged to website/ and ember-flight-icons/


amyrlam:
talked with @didoo and @MelSumner 

let's use a pre-push hook instead, that will lint on push

melanie shared https://github.com/okonet/lint-staged which i will look into!

_Originally posted by @amyrlam in https://github.com/hashicorp/flight/pull/233#discussion_r713247573_

amyrlam:
will move gears to the percy issue was my "side" issue and when start this again, can share a branch if is not working still.

amyrlam:
This will be done as part of the monorepo project.

amyrlam:
closing since will be addressed in the monorepo https://app.asana.com/0/1201630204387510/1201909305231864/f

PR #376, opened 2022-01-13
Try out yarn workspaces

amyrlam:
## :pushpin: Summary

Try out yarn workspaces for monorepo spike. Is hard to see the changes in editor so opening a WIP.

Followed these instructions: https://classic.yarnpkg.com/lang/en/docs/workspaces/
- Initially tried lerna, but some of the features don't work well with Yarn, see: https://github.com/lerna/lerna/issues/852
- Seemed easier to set up yarn workspaces than Turborepo for a first pass, but briefly tried yarn workspaces, lerna, and Turborepo

## :link: External Links

https://app.asana.com/0/1200695126769763/1201648749083347/f

***


amyrlam:
TODO:
- [ ] Figure out if the root package.json and having the npm workspaces include a slash e.g. `@hashicorp/flight-icons` is ok
- [ ] Try something like https://www.horacioh.com/writing/setup-lint-staged-on-a-monorepo next, as husky (for lint-staged) needs to be in the root. Related husky and lint-staged issue: https://github.com/hashicorp/flight/issues/258

.github/workflows/ci-ember-flight-icons.yml
```diff
@@ -32,7 +32,7 @@ name: CI - Ember Flight Icons
 
 defaults:
   run:
-    working-directory: ember-flight-icons
+    working-directory: @hashicorp/ember-flight-icons
```
amyrlam, line 35: @Dhaulagiri At a stopping point for the day. This change kills CI, something to debug tomorrow. Even in some other testing that I scrapped, not sure how to handle the package name including a `/`.

Am almost wondering if a better next step would be to bring 2 packages into a new monorepo, e.g. design-system-components and design-system-tokens. 

If you have any thoughts just lmk

amyrlam: TIL, thanks!

I'll find some examples of monorepos, and try to see if can debug the "/" in the package name

Dhaulagiri: can you say more about the `/` being a problem. is it a problem elsewhere besides the test here?

amyrlam: image

It seems solved here https://github.com/hashicorp/react-components#adding-package-dependencies but they're using npm workspaces and changesets. Still need to take a closer look into that example

Dhaulagiri:
For these experiments I would recommend forking the repo so that it a) makes less noise here in the production repo and b) frees you up to experiment more fully by merging to main and seeing what things look like outside of a branch which is unwieldy and hard to test certain things.

Issue #346, opened 2021-12-09
GitHub action doesn't trigger CI


Dhaulagiri:
https://github.com/hashicorp/flight/pull/322 created a github action for the build/sync process, but in practice the PR that is opened as a result of this is not correctly triggering the ci action to also run. This prevents the PR from being reviewed and merged.

We should either
* Ideally figure out why this is not working as expected and correct it
* or, less ideally, remove the action and associated documentation for it

Dhaulagiri:
[This is the reason](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs). The step that creates the PR can't trigger subsequent workflows like our CI run.

Dhaulagiri:
@didoo @MelSumner @amyrlam this is low priority, but if we want to be able to have this workflow work as expected we'll have to setup a personal access token in the action as outlined [here](https://github.com/peter-evans/create-pull-request/issues/48#issuecomment-536204092) which I can do but...Main drawback being the PR would be opened in that person's name and that person could not approve/request changes on the PR. And if that person disappeared from the company the token would also break.

👍  or 👎  any thoughts?

MelSumner:
@Dhaulagiri this is def not my domain of expertise, I usually rely on someone else on the team to handle these things. I will defer to the rest of the team in this case.

didoo:
my feedback is this: **in theory** using GH actions for automatically open PRs for the Flight icons is the perfect solution; **in practice** is solving a problem that probably we don't have, with all the drawbacks that we have seen. for me, we can even remove this automated task.

(one more useful would be to bump&release via single click/action; if only we could specify a parameter in the action which is the bumped version)

amyrlam:
If keep the action, start with you, but ask Helpdesk if they can create a "shared" user so token not tied to an individual? Wouldn't be surprised if there already was one or more at the company TBH

PR #377, opened 2022-01-13
Revert "create sync & build github action"

Dhaulagiri:
Fixes https://github.com/hashicorp/flight/issues/346

I'm still interested in fixing this so it works, but for now reverting makes sense given having this around is misleading.


didoo (approved):
Didn't test it, but LGTM :)

PR #382, opened 2022-01-14
Update "Vault" icon

### LINKED ISSUE #372: Flight Icon Request: [vault-color] (update)
**Outline the use case for the icon update.**
@ashleemboyer has highlighted a use case for the Vault (color) icon that is not supported by the current icon in Flight:

screenshot_695

Currently in Flight the `vault-color` icon is an outlier:
screenshot_696
but from the _brand_ color perspective we should use the yellow color for that icon
screenshot_697

Now, the reason for having the black colored version was that we were waiting for a redesign of the Vault logo by the marketing team. Now this redesign has happened (the difference is in the number of squares inside the logo) and it's available in the [02 - Global Assets v2.0](https://www.figma.com/file/aeLMZnYDsUAeEOUadjtjgP/02---Global-Assets-v2.0?node-id=2912%3A16277) file.

After quick chat with @dizzyup we have agreed that it's safe to update the `vault-color` icon in flight https://flight-hashicorp.vercel.app/?query=vault-color to use the brand color (yellow) and not the "alt" one. This will impact the IconTile component, but @dizzyup says it's OK.

screenshot_700

Please @heatherlarsen @cveigt can you update the icon in the Figma files, so we can release a new version? Thanks

didoo:
## :pushpin: Summary

This resolves #372.

In this PR I have:
- add the `animated` frame in the config of the build process (or the `sync` step would have not worked)
- regenerated the icons to include the updated Vault icon 
- bumped the package version to `2.0.2`


Dhaulagiri:
is it correct that `website/public/assets/zip/flight-icons-svg.zip` got deleted with this PR? (it's possible the script got broken during my website migration)

didoo:
> is it correct that `website/public/assets/zip/flight-icons-svg.zip` got deleted with this PR? (it's possible the script got broken during my website migration)

Mmm very strange. Let me rebuild (we had to select what to commit and maybe it was left out)

cveigt (approved):
LGTM! 👍

PR #392, opened 2022-01-24
Update documentation for React developers after feedback

didoo:
In [this Slack comment](https://hashicorp.slack.com/archives/C7KTUHNUS/p1643048914096100?thread_ts=1643048240.095000&cid=C7KTUHNUS) @ashleemboyer raises a couple of valid points:

- The description under the “React/SVG” heading is the same as the one under “Inline SVG”
- We’ve been importing the icons like this in dev-portal, which is different from what I’m seeing in this page. I tried copy/pasting what’s in the docs and I’m getting some errors. Will have to investigate a little further but need to join to a meeting in a few `import { IconArrowRight16 } from '@hashicorp/flight-icons/svg-react/arrow-right-16'`

In this PR I have fixed both issues.

Preview: https://flight-git-update-docs-for-react-devs-hashicorp.vercel.app/engineering#use-react

### DIFF
--- a/website/app/templates/engineering.hbs
+++ b/website/app/templates/engineering.hbs
@@ -329,11 +329,16 @@
     <CodeBlock
       @language="js"
       @code="
-          // import the SVG file
-          const flightArrowRight = require('@hashicorp/flight-icons/svg/arrow-right-24.svg');
+          // import the SVG file (using 'require')
+          const flightArrowRight = require('@hashicorp/flight-icons/svg/arrow-right-24.svg?include');
+          // or import the SVG file (using 'import')
+          import flightArrowRight from '@hashicorp/flight-icons/svg/arrow-right-24.svg?include';
 
           // elsewhere in the file
           <InlineSvg src={flightArrowRight} />
+
+          // alternatively you can also use a similar approach
+          <InlineSvg src={require('@hashicorp/flight-icons/svg/arrow-right-24.svg?include')} />
         "
     />
     {{! prettier-ignore-end }}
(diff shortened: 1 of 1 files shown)


website/app/templates/engineering.hbs
```diff
@@ -355,29 +360,19 @@
...
     {{! prettier-ignore-start }}
     <CodeBlock
       @language="js"
       @code="
-          // import the React file (TypeScript)
-          const FlightArrowRight = require('@hashicorp/flight-icons/svg-react/arrow-right-24');
+          // import the React/TypeScript file (using 'require')
+          const IconArrowRight24 = require('@hashicorp/flight-icons/svg-react/arrow-right-24?include');
+          // or import the React/TypeScript file (using 'import')
+          import { IconArrowRight24 } from '@hashicorp/flight-icons/svg/arrow-right-24.svg?include';
```
ashleemboyer, line 372: Noticed a few things here:

- the `?include` and `.svg?include` shouldn't be in the require/important statements
- I needed to put curly braces around `IconArrowRight24` for the `require` statement to work
- the import statement needs to be updated to be from `@hashicorp/flight-icons/svg-react` instead of `@hashicorp/flight-icons/svg`

```suggestion
          // import the React/TypeScript file (using 'require')
          const { IconArrowRight24 } = require('@hashicorp/flight-icons/svg-react/arrow-right-24');
          // or import the React/TypeScript file (using 'import')
          import { IconArrowRight24 } from '@hashicorp/flight-icons/svg-react/arrow-right-24';
```

ashleemboyer (changes requested):

website/app/templates/engineering.hbs
didoo: @ashleemboyer thanks for the suggestion/fix.

regarding the `?include` I followed in the import/require statements in the DevPortal codebase:
https://github.com/hashicorp/dev-portal/blob/d82ace559de1226bf3d75c72197af6de03b5a1ed/src/components/_proxied-dot-io/waypoint/homepage/sections/how-it-works/build-and-deploy/index.tsx#L76
https://github.com/hashicorp/dev-portal/blob/1a000fa43c9fdeed2e90f1a579b0c251c7663246/src/components/navigation-header/index.tsx#L3

ashleemboyer: No problem! 😊 So the `?include` in the import statement works when we're importing an svg file and using the `InlineSvg` component, but not when we're importing React components.

ashleemboyer (approved):

website/app/templates/engineering.hbs
didoo: Ohhhh! I see :) good to know!

MelSumner (approved):
:shipit:

PR #393, opened 2022-01-27
chore: Update structure mappings to correct missing icons

johncowen:
Note: I wasn't quite sure who to add as reviewer, please feel free to add any other relevant folks!

## :pushpin: Summary

Updates the structure-mappings file to correct missing icons

N/A

## :camera_flash: Screenshots

N/A

## :link: External Links

N/A

***


didoo:
@heatherlarsen @cveigt assigned to you for review (and removed me and Hector, it's just a matter of checking the mapping).

structure-mappings.json
```diff
@@ -151,8 +151,8 @@
   "radio-button-unchecked": "circle",
   "random": "random",
   "redirect": "redirect",
-  "refresh-alert": "refresh-alert",
-  "refresh-default": "reload",
+  "refresh-alert": "sync-alert",
+  "refresh-default": "sync",
```
heatherlarsen, line 155: "reload" is actually the correct mapping here. 
Structure Refresh - Default: https://github.com/hashicorp/structure-icons/blob/master/src/Refresh%20-%20Default.svg
Flight Reload: https://flight-hashicorp.vercel.app/?query=reload

heatherlarsen (changes requested):
@johncowen Found one that needs updating, otherwise these are great changes. Thanks for doing this!

structure-mappings.json
johncowen: Ah thanks Heather! Fixed that up

---

I spotted a few more that where missing also (just 3 brand icons), shall I just pop one more commit on the end of this PR, then you can check that commit also before merge (might be easier than making a fresh PR just for that?)

structure-mappings.json
```diff
@@ -151,8 +151,8 @@
   "radio-button-unchecked": "circle",
   "random": "random",
   "redirect": "redirect",
-  "refresh-alert": "refresh-alert",
-  "refresh-default": "reload",
+  "refresh-alert": "sync-alert",
+  "refresh-default": "refresh",
```
heatherlarsen, line 155:
```suggestion
  "refresh-default": "reload",
```

structure-mappings.json
johncowen: Doh 🤦 thanks for the suggestion reload/refresh 😆  more or less the same 😬 😄

johncowen:
K those extra ones are there @heatherlarsen whenever you get a sec, FYI I'm not blocked by this so no rush

johncowen:
Hey @heatherlarsen

I popped another commit on the end of here (sorry I know some folks prefer adding commits, and some prefer new PRs but figured for this new PR would be overkill)

@jnwright noticed that the mapping here for Structure's triangle (which by default is filled) is wrong as it points to Flights outlined variant. The triangle in the following image should be filled:

image (2)

I updated the mapping and checked it locally:

Screenshot 2022-02-03 at 09 17 11

I also noticed that the circle one here is probably wrong also, and I'm guessing there are possibly a few more, do you want me to just keep pushing on the end of here, or shall I make new PRs when we spot them?

ty!

heatherlarsen (approved):
lgtm! :shipit:

heatherlarsen:
@johncowen adding here is ok with me.

Dhaulagiri:
I'm going to bring this in as-is for now. If there are additional changes feel free to send them along.

PR #398, opened 2022-02-02
Bump “flight-icons” dependency in website

didoo:
## :pushpin: Summary

In this PR I have bumped the “flight-icons” dependency in the `website` folder, to trigger a re-build of the website and show the newly released icons


Dhaulagiri:
I mistakenly approved the percy diff when it's not actually correct. Will try to 👀  today to see how to get the packages to be in the correct place.

Dhaulagiri:
Whatever dependabot did in #400 seems to have fixed this more properly, so I think we can close this one (would be helpful to figure out how to do the same thing manually though)

PR #406, opened 2022-02-10
Update icon request template

heatherlarsen:
Merging this PR would remove @heatherlarsen from the assignee list in the icon request template. @cveigt will be managing these requests moving forward and will assign to me as he sees fit based on team bandwidth.


didoo (approved):
👍

Issue #407, opened 2022-02-11
Locally linking ember-flight-icons in website doesn't work


Dhaulagiri:
### Steps to Reproduce
```bash
cd ember-flight-icons
yarn link
cd ../website
yarn link @hashicorp/ember-flight-icons
yarn start
```

### What happens
<img width="918" alt="Screen Shot 2022-02-11 at 08 26 37" src="https://user-images.githubusercontent.com/1672302/153609202-0784f2e9-ee8d-4bb5-b202-ff7f1c7241c1.png">

### What I expect
The website load as normal

Dhaulagiri:
Note: this issue came up trying to experiment with these packages in a monorepo where linking is required for things to work. Hence a higher priority to figure out.

didoo:
Yes, I saw the same error myself, while doing some tests, but I couldn't figure out what was the cause. Great to see it fixed :)

PR #408, opened 2022-02-11
Update ember-flight-icons to ember-source 4.1

Dhaulagiri:
## :pushpin: Summary

* npx ember-cli-update
* accept the changes to our ci config (will drop if they prove tricky to get working)

Fixes https://github.com/hashicorp/flight/issues/407 (have to test it locally to confirm which I did)

Unsure why, but presumably there was some bug in the version of Ember we were using in ember-flight-icons

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1201824169692360


MelSumner (approved):
:shipit: Noting for any observers not familiar with Ember's major version policy that major versions of Ember only remove deprecated code; they don't introduce breaking changes or new features.

PR #410, opened 2022-02-11
Test cleanup

Dhaulagiri:
## :pushpin: Summary

* Remove ci file accidentally committed in #408 
* The concurrency setup was causing 1 of the jobs to get cancelled which is not what we want. This limits concurrency/cancellation to each job individually.


MelSumner (approved):
:shipit:

PR #421, opened 2022-03-04
Delete files

amyrlam:
## 📌 in: Summary

We will deprecate the repo (moved to `design-system`) but this change could make it clearer in the UI


amyrlam:
Let's leave it as-is for now

